PR #55, opened 2014-12-26
More sign in tweaks

theengineear:
Some tweaks to get our online docs the way we want wrt sign-in/authentication info.


theengineear:
@etpinard and @chriddyp , made the above changes and am merging in, fyi.

PR #67, opened 2015-05-20
Create 2015-05-20-horizontal-bar

aneda:
Should we keep everything the same style?  for example here defining trace separately as an object


chriddyp:
yeah, i think that's right

PR #70, opened 2015-05-20
Create 2015-05-20-marker-bar

aneda:


chriddyp:
:dancer: that opacity + border line looks great!

PR #74, opened 2015-05-20
Create 2015-05-20-text-with-annotations

aneda:


_posts/python/bar/2015-05-20-text-with-annotations
```diff
@@ -0,0 +1,120 @@
...
+                xref= 'x',
+                yref= 'y',
+                xanchor='center',
+                yanchor='top',
+                text='January 2013 Sales',
+                font=Font(
+                    family='Arial',
+                    size=20,
+                    color= 'rgb(8,48,107)',
+                ),
```
chriddyp: since this example is getting so long, we might want to remove the custom `font` and introduce it in the separate customizing `annotation`s chapter, and include *just* the necessary code to add text annotations on top of the bars for this example

_posts/python/bar/2015-05-20-text-with-annotations
```diff
@@ -0,0 +1,120 @@
...
+        showticklabels = False,
+    ),
+    
+  annotations = Annotations([   
+            
+        Annotation(
+                x= 'Product A',
+                y= 30,
+                xref= 'x',
+                yref= 'y',
```
chriddyp: `xref` and `yref` might be `x` and `y` by default, in which case we can remove them from here

_posts/python/bar/2015-05-20-text-with-annotations
```diff
@@ -0,0 +1,120 @@
...
+                xanchor='center',
+                yanchor='top',
```
chriddyp: what happens if we make this `bottom` and change `y` to be `20` (the value of the y-data). It *might* sit on top of the bar OK then, which would make this a bit more extensible for folks using different sets of data.

_posts/python/bar/2015-05-20-text-with-annotations
```diff
@@ -0,0 +1,120 @@
...
+        zeroline = False, 
+        showline = False,
+        showticklabels = False,
+    ),
+    
+  annotations = Annotations([   
+            
+        Annotation(
+                x= 'Product A',
+                y= 30,
```
chriddyp: what happens if we change this `y` to be `30` (the value of the y-data) and then change `yanchor` to `bottom`? It *might* sit on top of the bar OK then, which would make this a bit more extensible for folks using different sets of data. If so, then we could modify the example to initialize the `y` list at the top, and reference it here as `y[0]`

_posts/python/bar/2015-05-20-text-with-annotations
```diff
@@ -0,0 +1,120 @@
...
+        x=['Product A', 'Product B', 'Product C'],
+        y=[20, 14, 23],
+        marker = Marker(
+                color = 'rgb(158,202,225)',
+                opacity = 0.6,
+                line =Line(
+                    color ='rgb(8,48,107)',
+                    width = 1.5
+                ),
+        ),
```
chriddyp: if we're *just* going to be introducing `text` on top of bars for this section, we *might* want to keep this part really simple, and remove the `marker` part (even though it makes the example look so much nicer)

aneda:
@chriddyp it's ready!

chriddyp:
nice! :dancer:

PR #84, opened 2015-05-26
Create 2015-05-26-gallery-time-series

aneda:
@chriddyp  I am not sure about this one.  I think we can replace it with something nicer.  For this data set if we present the x-axis with only %d format (instead of %m%d) , it would look nicer and less crowded:
https://plot.ly/~neda/2522/montreal-temperature-february-2015/


chriddyp:
yeah, I think we could probably replace it w something nicer. The tick format is great though, and maybe we can refactor this example into the time-series section?

PR #96, opened 2015-05-27
Migrate python user guide notebooks over

chriddyp:
hey @NicoleGrondin -- here's a script that I wrote and used to port these notebooks over: http://nbviewer.ipython.org/gist/chriddyp/cb693d3ca619ed96a59c

they already exist in this repo: github.com/plotly/python-user-guide, that script basically just copies them over and adds some new front matter

i also grabbed some css that we were using in the main site for the notebooks and included them in here, too.


_includes/head.html
```diff
@@ -1,27 +1,34 @@
...
+    	<link rel="stylesheet" type="text/css" href="{{site.staticurl}}/css/base.css">
+    	<link rel="stylesheet" type="text/css" href="{{site.staticurl}}/css/layout_block.css">
+    	<link rel="stylesheet" type="text/css" href="{{site.staticurl}}/css/modules.css">
+    	<link rel="stylesheet" type="text/css" href="{{site.staticurl}}/css/state.css">
+
+        {% if page.page_type == "ipython" %}
+        <link rel="stylesheet" type="text/css" href="{{site.staticurl}}/css/ipython_notebook.css">
+        {% else %}
+        <link rel="stylesheet" type="text/css" href="{{site.staticurl}}/css/api_docs.css">
+        {% endif %}
```
chriddyp: @NicoleGrondin new `page_type`! kind of a cool way to add conditionals backwards from the post front matter up to the original templates

chriddyp:
they look *pretty* good! i haven't looked through all of them yet, and I bet we'll need to make some tweaks and update some urls etc. I'll merge this in for now, though, so that we can play around w it

image

PR #99, opened 2015-05-29
Create 2015-05-29-box-outliers

aneda:


aneda:
@chriddyp

chriddyp:
@aneda and then close this PR and just use the 4 box plots from the other example?

PR #100, opened 2015-05-29
Create 2015-05-29-box-suspected-outliers

aneda:


aneda:
@chriddyp is this what you had in mind for suspected outliers?

_posts/python/box/2015-05-29-box-mean
```diff
@@ -0,0 +1,33 @@
...
+    y = [2.37, 2.16, 4.82, 1.73, 1.04, 0.23, 1.32, 2.91, 0.11, 4.51, 0.51, 3.75, 1.35, 2.98, 4.50, 0.18, 4.66, 1.30, 2.06, 1.19],
+    name = 'Mean & SD',
+    marker = Marker(
+         color = 'rgb(10, 140, 208)',
+    ),
+    boxmean = 'sd'
+    
+)
+data = Data([trace0, trace1])
+plot_url = py.plot(data, filename='basic-box-plot-mean')
```
chriddyp: woops, looks like I merged this one from the previous PR. @aneda  could you delete this example from this PR?

_posts/python/box/2015-05-29-box-suspected-outliers
```diff
@@ -0,0 +1,30 @@
+---
+name: Box Plot Styling Suspected Outliers 
+plot_url: https://plot.ly/~neda/2747/suspected-outliers/
+language: python
+suite: box
+order: 7
```
chriddyp: we might want to include a "description: " variable with something along the lines of:

> In this dataset, 0.75 and 16 are both "suspected" outliers and 20.90, 22.30, and 23.25 are actual outliers.

_posts/python/box/2015-05-29-box-suspected-outliers
```diff
@@ -0,0 +1,30 @@
...
+# Learn about API authentication here: https://plot.ly/python/getting-started
+# Find your api_key here: https://plot.ly/settings/api
+
+import plotly.plotly as py
+from plotly.graph_objs import *
+
+trace0 = Box(
+    y = [0.75, 5.25, 5.5, 6, 6.2, 6.6, 6.80, 7.0, 7.2, 7.5, 7.5, 7.75, 8.15, 8.15, 8.65, 8.93, 9.2, 9.5, 10, 10.25, 11.5, 12, 16, 20.90, 22.3, 23.25],
+    name = 'Suspected Outliers',
+    boxpoints = 'suspectedoutliers',
```
chriddyp: @aneda maybe we should include another example with 'outliers' but the same dataset. and then maybe replace use this same data for [this example](https://github.com/plotly/documentation/blob/gh-pages/_posts/python/box/2015-04-09-box-plot-jitter.html) ? that way folks can see the same dataset with all the points, with just outliers, with outliers + suspected outliers.

chriddyp: @aneda woops, nvm, sorry didn't see this PR https://github.com/plotly/documentation/pull/99/files yet. what do you think about including a 3rd box in that example: https://github.com/plotly/documentation/pull/99/files with `suspectedoutliers` and then removing this example?

aneda:
@chriddyp is this what you have in mind 
https://plot.ly/~neda/3455/only-whiskers-suspected-outliers-whiskers-outliers/
I like to have all there in one

chriddyp:
looks good! should we add a fourth box that shows `all` of the points?

aneda:
@chriddyp like this?
 https://plot.ly/~neda/3455/all-points-only-whiskers-suspected-outliers-whiskers-outliers/
visually is good and the code is still on the short side, but would it be too much new info at once?

chriddyp:
yeah! I think that looks great. I think it's helpful to see the original data next to the boxes with the outliers. Good point re too much new info at once... maybe let's place this example after the readers get warmed up with "Box Plot That Displays the Underlying Data"?

aneda:
@chriddyp sorry for the delay
This example would be part of Advanced section so I think users are warmed up.  We can replace all with one (https://plot.ly/~neda/3455/all-points-only-whiskers-suspected-outliers-whiskers-outliers/)  and if later on we thought we need more warm up, we'll change it.  Is that good?  Should I replace the code?

chriddyp:
yeah, let's replace all with https://plot.ly/~neda/3455, I love that example

PR #102, opened 2015-06-05
Create 2014-06-05

aneda:
@chriddyp  I can also make something like this but smaller so it fits 600 x 600:
https://plot.ly/~neda/2914


chriddyp:
looks good!

PR #113, opened 2015-06-18
Create 2015-06-18-bubble-marker-size

aneda:
@chriddyp is this good?  (introducing size in Marker for bubble charts)


chriddyp:
:dancer:

PR #114, opened 2015-06-18
Create 2015-06-18-bubble-color

aneda:
@chriddyp

### DIFF
--- a/_posts/python/bubble/2015-06-18-bubble-color
+++ b/_posts/python/bubble/2015-06-18-bubble-color
@@ -9,24 +9,20 @@ import plotly.plotly as py
 from plotly.graph_objs import *
 
 trace0 = Scatter(
-    x = [1, 2, 3, 4],
-    y = [10, 11, 12, 13],
-    mode = 'markers',
-    marker = Marker(
-        color = ['rgb(93, 164, 214)', 'rgb(255, 144, 14)',  'rgb(44, 160, 101)', 'rgb(255, 65, 54)'],
-        opacity = [1, 0.8, 0.6, 0.4],
-        size = [40, 60, 80, 100],
+    x=[1, 2, 3, 4],
+    y=[10, 11, 12, 13],
+    mode='markers',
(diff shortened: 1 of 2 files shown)


chriddyp:
:dancer: , and let's remove this one: https://github.com/plotly/documentation/blob/gh-pages/_posts/python/bubble/2015-04-09-bubblechart.html now that we have these. 
we should also probably include a `color` example that uses numerical values and the colorscale instead of just color strings

_posts/python/bubble/2015-06-18-bubble-color
```diff
@@ -0,0 +1,32 @@
+---
+name: Marker Size & Color Styling for Bubble Chart
```
chriddyp: maybe just `Setting Marker Size and Color` since we're already in the bubble chart section?

aneda:
@chriddyp  do you mean something like this https://plot.ly/3730/~etpinard/ 
this has both colorscale and color as number
ps. I merged the bubble-text here and will correct it later

chriddyp:
@aneda yeah, exactly

aneda:
@chriddyp ok sounds good.   I think we should delete this one as well:
https://github.com/plotly/documentation/blob/gh-pages/_posts/python/bubble/2015-04-09-bubble-chart-sizeref-sizemode.html

chriddyp:
@aneda that sounds good, I'd still love to see a simple illustrative example of how `sizeref` actually works

aneda:
@chriddyp you mean another simple example for sizeref beside this one:

image

https://github.com/plotly/documentation/blob/gh-pages/_posts/python/bubble/2015-06-18-bubble-marker-size

chriddyp:
ah nvm, I missed that sizeref one. cool, good to remove https://github.com/plotly/documentation/blob/gh-pages/_posts/python/bubble/2015-04-09-bubble-chart-sizeref-sizemode.html? and merge?

PR #121, opened 2015-06-18
Create 2015-06-18-box-grouped-horizontal

aneda:
@chriddyp do we want to add this or not?  the only addition is: * orientation = 'h' *


chriddyp:
yeah, I think that's great. :dancer:

PR #132, opened 2015-06-23
Create 2015-06-23-contour-smoothing

aneda:
@chriddyp


chriddyp:
@aneda - it took me a while to figure out what smoothing was doing. Maybe this example should use `make_subplots` and have identical contour plots, one with smooth lines and the other without smoothing. We also might want to use a dataset that has more extreme contour angles.

PR #140, opened 2015-06-23
IDs for sub-example links

NicoleGrondin:
added id so that all sub examples can be linked# id lowercased and spaces replaced by dash
format: sub-example-title

@chriddyp  not sure which branch we're running things on these days...

### DIFF
--- a/_includes/auto_examples.html
+++ b/_includes/auto_examples.html
@@ -1,26 +1,30 @@
 {% for example in examples %}
     <div class="section">
         <div class="row auto-eg-padding">
-			<div class=" row twelve columns">
-                <h3>{{ example.name }}</h3>
-			</div>
-			<div class="row">
-				<div class="six columns">
-					<div class="z-depth-1">
-						<pre><code class="{{page.language}}">{{ example.content }}</code></pre>
-					</div>
-					{% if example.description %}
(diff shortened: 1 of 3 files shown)


_includes/auto_examples.html
```diff
@@ -1,26 +1,30 @@
...
-				</div>
-				<div class="six columns">
-					<div class="border">
-						<iframe src="{{ example.plot_url }}.embed" style="width: 100%; height: 550px; border: none;"></iframe>
-					</div>
-				</div>
-			</div>
+      <div class=" row twelve columns">
+        {% if example.name contains '<code>'%}
+          <h3 id="{{ example.name | remove: '<code>' | remove: '</code>' | replace:' ','-' | downcase }}">{{ example.name }}</h3>
```
chriddyp: instead of dynamically modifying the code tags we should probably just remove them from the examples with a big find-and-replace.

chriddyp: instead of `remove: '<code>' | remove: '</code>'`: ` | striptags` to handle any other tags that we might stick in there (http://jinja.pocoo.org/docs/dev/templates/#striptags)

chriddyp:
other than my comment, :dancer:

PR #147, opened 2015-06-24
Create 2015-06-24-multiple-make-multiple-specs

aneda:
@chriddyp


aneda:
@chriddyp  I am introducing specs with selective titles for subplots.  Should we have them as seperate examples or is it ok?

chriddyp:
this looks good. I think the first example should be a simple subplot w/out subplot titles, the second example should be the same example but w/subplot titles, and the rest of them can be w/out titles (up to your judgement)

PR #151, opened 2015-06-29
publish notebooks in less time than it takes to say "nbviewer"

chriddyp:
@etpinard @NicoleGrondin @aneda @theengineear @mkcor 

here's a new approach to publishing of notebooks to plotly. the workflow:

1 - write your ipython notebook. this may be an in-depth data science tutorial (e.g. https://plot.ly/ipython-notebooks/big-data-analytics-with-pandas-and-sqlite/), a quick answer to someone's question (e.g. http://nbviewer.ipython.org/gist/chriddyp/833e11ffe240ef49a463), or some explaining the basics of python client (e.g. https://plot.ly/python/file-options/)

2 - include a code block at the end of the notebook that converts the current notebook that you are writing into a valid gh-pages html document

3 - push the notebook and the html file up to this repo

Step 2 is the alternative to including external config files and running python scripts. There is a real messy 4 blocks of code in the notebook in this PR that converts itself into a valid gh-pages html file (https://github.com/plotly/documentation/blob/faster-than-nbconvert/_posts/python/scattergeo/NWS-Rainfall-Telemetry.ipynb, cells 13-18). We could turn this into an importable single line of code though, something like:

```
import nbtoghpages
nbtoghpages(filename, frontmatter)
```

image

to make this really solid, we'll need a couple of things:
- [ ] a new template, just like the python user guide template, but that doesn't have the "Plotly for Python - User Guide" header
- [ ] better notebook css. I'd really like to move all of the notebooks over to this CSS: https://raw.githubusercontent.com/plotly/python-user-guide/css-updates/custom.css, but we'll need to make a couple modifications to the classes so that they don't conflict with the universal header
- [ ] Make Step 2 super easy and importable


chriddyp:
nbconversion step: 
image

etpinard:
@chriddyp This is definitely the way to go. `make` scripts are out.

That said, there are a couple of steps currently part of the publish process that would be nice to keep (if not too much of a hassle):

- [update the anchors](https://github.com/plotly/IPython-plotly/blob/master/_makescripts/publish.py#L69) so that outbound links get a `target="_blank"`
- [add anchors to In/Out code cells](https://github.com/plotly/IPython-plotly/blob/master/_makescripts/publish.py#L87) which is very useful when sharing content
- [convert the nb to plain python](https://github.com/plotly/IPython-plotly/blob/master/makefile#L45) and a link in the NB header - for old school folks.

And perhaps we should also [wget the images to serve them](https://github.com/plotly/IPython-plotly/blob/master/_makescripts/publish.py#L16) ?

chriddyp:
in action: https://github.com/plotly/documentation/pull/156

chriddyp:
basically closed with https://github.com/plotly/documentation/pull/165, a few outstanding issues i'll add in sep issues

PR #167, opened 2015-07-14
Migrate notebooks

chriddyp:
how fun!

ipython notebook migration part 1

@jackparmer - these notebooks use this css: https://github.com/plotly/documentation/blob/gh-pages/all_static/css/userguide_style.css, and it could use some seriously love. 

these notebooks are in a `tutorials` folder, along with their HTML version. check out the HTML front-matter to see the URLS. they are currently the same as on plot.ly, but instead of /ipython-notebooks/, it's at /python/, e.g. plot.ly/python/big-data-analytics-with-pandas-and-sqlite/

once the css is cleared up i'll move them back to /ipython-notebooks/ and wire it up with the new python-notebooks index page


jackparmer:
Hey Chris, try this css sheet - http://bit.ly/1O4UAe6

PR #177, opened 2015-08-18
Header consistency

NicoleGrondin:


chriddyp:
what's the difference between first-header.html and header.html? is it necessary to have two of them?

NicoleGrondin:
@chriddyp now simplified down to one header. The splash page used to have a different header but they're the same anyway now.

chriddyp:
:dancer:

PR #209, opened 2015-11-13
Adding codepens

tarzzz:
Updated the Codepen embed size for the documents


tarzzz:
This should be compared against 'source' branch. Closing and raising another PR

PR #237, opened 2015-12-03
Fix a small typo.

hartzell:
I think that this should read:

"Now you can pass .... either **or** the ...", not "either **on** the...".


mdtusz:
Yep! Good catch. 

:dancer:

PR #263, opened 2015-12-27
Added 'arrangement: horizontal' to all the plots

tarzzz:
@jackparmer I was hoping to get a quick review/go-ahead before I merge this change. It is a big change, and spans about 700 files. Thanks!


jackparmer:
OK, looks good to me! Let's merge it!

PR #274, opened 2016-01-07
WIP: jsdocs pagespeed improvements

mdtusz:
The page will now load in ~20s instead of ~35.42s. The major bottleneck is the turbulence simulation plot - I will try reducing the number of data points it needs to chew through, but if that still isn't sufficient, we should either use an image for that plot, or choose a new, less data heavy plot.


mdtusz:
I added a pretty basic scrollspy that will load the plots as they scroll into view, but that turbulence simulation plot is just slowwww (understandably from the amount of data it uses).

Can we use a different plot as an example? I tried reducing the amount of data by half, but that's not a very good solution - it was still quite slow and really makes the plot look incomplete.

chriddyp:
@mdtusz let's replace it with an image for now, and a caption to view the full screen interactive plot

mdtusz:
:+1: sounds good to me.

mdtusz:
@jackparmer if you're happy with the changes I'll merge this in.

jackparmer:
Thanks @mdtusz for tackling this! Merge away!
We get >60k uniques to this page per month, so this will make a lot of visitors much happier.

mdtusz:
:+1: 

We can still make improvements to it, but this will be a good start.

PR #325, opened 2016-02-22
Add a note about plotly.offline.plot

ischurov:
It seems that `plotly.offline.plot` is underdocumented: at least I wasn't able to find any mentions of it before investigating the code of plotly.offline. I believe that we should mention it somewhere. This file seems to be appropriate place despite the fact it is devoted mostly to `iplot`.


cldougl:
Thanks for the contribution- however plotly offline documentation has been added [here](https://plot.ly/python/offline/) as well as in the getting started page: https://plot.ly/python/getting-started/#initialization-for-offline-plotting!

PR #374, opened 2016-04-28
Ternary contour

tarzzz:
@cldougl : The ternary plot is rendering correctly (w/o any errrors) but the fill-colors are not showing up on the plot.


tarzzz:
Merging!!

PR #394, opened 2016-05-30
Add missing array terminator

adamlwgriffiths:
The visible example code seems to have lost the closing bracket of the array.
This commit re-adds it.


yankev:
👍  Thanks.

PR #525, opened 2016-09-15
Add optional markdown block to auto examples

rreusser:
This PR allows frontmatter to contain markdown content that prefixes examples. Usage is:

```yaml
---
title: Foo
markdown_content: |
  ### This a heading, though you probably shouldn't use it in the markdown

  This is a paragraph

  This is another paragraph
---
```

A general principle (via @cldougl) is to keep the examples simple and *avoid* the need for lots of explanation, but the animation docs basically require some additional paragraph-formatted content. This works it into the example files without the need for larger structural changes (like a custom layout?)

### DIFF
--- a/_includes/auto_examples.html
+++ b/_includes/auto_examples.html
@@ -7,6 +7,9 @@
       </div>
       <div class="row">
         <div class="{% if example.arrangement contains 'horizontal' %}twelve{% else %}six{% endif %} columns">
+          {% if example.markdown_content %}
+            {{ example.markdown_content | markdownify }}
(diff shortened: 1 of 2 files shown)


cldougl:
Awesome addition @rreusser !
Feel free to merge 💃

rreusser:
@cldougl Pardon the late addition, but I've added a bit of explanation and minor markdown cleanup to the JS readme. If it looks good, will merge.

cldougl:
Thanks for even updating the README @rreusser !!!

PR #528, opened 2016-09-15
Add peak integration

Kully:
cc @cldougl


cldougl:
@Kully you're trying to merge into master

PR #532, opened 2016-09-20
Upgrade jekyll and switch to Gemfile-based bundling

rreusser:
This PR updates `jekyll` to `3.2.1`, gets rid of `gem install` commands, and moves the dependencies into the Gemfile. This makes it easier for collaborators to execute `bundle install` and then `bundle exec ...` with the correct version of commands since the installs are no longer global.

`Gemfile.lock` has a few more changes than I'd like, so this should be checked carefully before committing to this, but I've confirmed that the dev server works, so hopefully things are… okay?


cldougl:
much needed @rreusser !!!
I've completed/updated this in different pr and updated [instructions](https://github.com/plotly/documentation/blob/source-design-merge/Contributing.md) and [js specific instructions](https://github.com/plotly/documentation/blob/source-design-merge/_posts/plotly_js/README.md)  (while setting up my new machine) so I'll close this up

PR #549, opened 2016-10-05
Removing redundant and difficult to parse header descriptions

fresheneesz:
The "DO THIS WITH plotly.whatever" looks terrible. Removing the redundant "DO THIS WITH" part and removing the html code nodes so that the function names are properly front and center.

### DIFF
--- a/_posts/2016-06-03-plotly_js_function_ref.html
+++ b/_posts/2016-06-03-plotly_js_function_ref.html
@@ -10,7 +10,7 @@ redirect_from: /javascript-graphing-library/plotlyjs-function-reference
 
 <h1>plotly.js Function Reference</h1>
 
-<h4 id="retrieving-data-layout"><a class="no_underline plot-blue" href="#retrieving-data-layout">Retrieving the plot <code>data</code> or <code>layout</code></pre> objects</a></h4>
+<h4 id="retrieving-data-layout"><a class="no_underline plot-blue" href="#retrieving-data-layout">plotNode.data <code>and</code> plotNode.layout</a></h4>
 
 The plot <code>data</code> or <code>layout</code> can  be retrieved from the <code>&lt;div&gt;</code> element in which the plot was drawn:
 
@@ -22,7 +22,7 @@ var plotDiv = document.getElementById('examplePlot');
 var plotData = plotDiv.data;
 </code></pre>
 
-<h4 id="plotly-newplot"><a>Plot with <code>Plotly.newPlot</code></a></h4>
+<h4 id="plotly-newplot"><a>Plotly.newPlot</a></h4>
 
 Use <code>newPlot</code> to create a new plot in an empty <code>&lt;div&gt;</code> element.
 A note on sizing: You can either supply height and width in <code>layout</code>, or give <code>graphDiv</code> a height and width in css.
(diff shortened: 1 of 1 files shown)


_posts/2016-06-03-plotly_js_function_ref.html
```diff
@@ -10,7 +10,7 @@
 
 <h1>plotly.js Function Reference</h1>
 
-<h4 id="retrieving-data-layout"><a class="no_underline plot-blue" href="#retrieving-data-layout">Retrieving the plot <code>data</code> or <code>layout</code></pre> objects</a></h4>
+<h4 id="retrieving-data-layout"><a class="no_underline plot-blue" href="#retrieving-data-layout">plotNode.data <code>and</code> plotNode.layout</a></h4>
```
cldougl: If you'd like to streamline this title, I'd suggest perhaps:
Retrieve data or layout objects.
I'm not sure 'plotNode.data and plotNode.layout' improves this title, as that's not the syntax used in the below example.

Furthermore, I understand why you dislike the <code> text in the titles and agree that can be removed, but I would *not* apply code syntax to non code-based words.

_posts/2016-06-03-plotly_js_function_ref.html
```diff
@@ -75,7 +75,7 @@ <h4 id="plotly-newplot"><a>Plot with <code>Plotly.newPlot</code></a></h4>
 
 There are several other options that you can supply as the fourth argument. See more examples of the <a href="https://plot.ly/javascript/configuration-options/">configuration options.</a>
 
-<h4 id="plotly-restyle"><a class="no_underline plot-blue" href="#plotly-restyle">Edit attributes with <code>Plotly.restyle</code></a></h4>
+<h4 id="plotly-restyle"><a class="no_underline plot-blue" href="#plotly-restyle">Plotly.restyle</a></h4>
 
 A more efficient means of changing parameters in the data array. When restyling, you may choose to have the specified changes effect as many traces as desired. The update is given as a single object and the traces that are effected are given as a list of traces indices. Note, leaving the trace indices unspecified assumes that you want to restyle <b>all</b> the traces.
```
cldougl: since editing attributes is removed from the title, it would be nice to add it into the text below.

_posts/2016-06-03-plotly_js_function_ref.html
fresheneesz: Which do you think would make more sense, changing the example syntax to `plotNode.data` etc or change the title to `plotDiv.data`?

I would not apply code syntax to non code-based words. I agree in general, but I did it so that those words would be deemphasized. Is there another way to do that with the styling these docs use?

_posts/2016-06-03-plotly_js_function_ref.html
fresheneesz: I figured "changing parameters" mean the same thing. Does it not?

cldougl: Yep- it was written to paraphrase the title, however, in the library and elsewhere in the documentation we typically refer to parameters as 'attributes' rather than parameters. I could have been more clear in my comment!! My point was since 'attributes' was the preferred phrasing, I'd like to keep that in this example (and the changing parameters line can be removed.)

_posts/2016-06-03-plotly_js_function_ref.html
cldougl: I'd vote for `plotDiv.data` because it is found in more examples in other areas of the documentation.

Unfortunately, we don't have a deemphasis style option for titles right now.

fresheneesz: Ok, I suppose a single 'and' in there won't be too distracting.

_posts/2016-06-03-plotly_js_function_ref.html
fresheneesz: Ok

fresheneesz:
@cldougl I fixed the things you mentioned

cldougl:
@fresheneesz awesome- thanks! Feel free to merge! 
Were you able to check out the instructions on deploying as well? If not I can do that step for you.
Not sure if you saw: https://github.com/plotly/documentation/blob/source-design-merge/_posts/plotly_js/README.md but it's a bit more thorough (js doc specific) than the contributing page.

fresheneesz:
"Feel free to merge!"

I don't have access to merge if you were talking to me. "Only those with write access to this repository can merge pull requests." I assume the same thing prevents me from deploying. 

I hadn't seen that readme page. Perhaps it would make sense to link to there from the Contributing.md page?

PR #643, opened 2017-01-11
Change heatmap animation permalink

Kully:


Kully:
@cldougl Changing permalink from `heatmap` to `heatmap-animation` since `heatmaps` already exists for just the Heatmaps page.

cldougl:
💃

PR #655, opened 2017-01-17
Python data upload styling

Kully:


Kully:
@cldougl So when I switched the base branch to `source-design-merge`, it said that both branches were identical and so I couldn't open the PR.

I was on a separate branch locally so I don't know why this happened.

PR #675, opened 2017-01-30
added delete-all-plots-and-grids example

Kully:
Ready for a review. I just added the delete-all-plots-and-grids to the bottom of the privacy doc page.
cc @cldougl


Kully:
@cldougl Okay, now it's looking good. Only thing I didn't do is update the linked `ipython notebook` for download. I can update after this gets merged as I don't want to keep the user hanging much longer.

cldougl:
💃

PR #700, opened 2017-02-23
Add docs and a codepen explaining null vs. undefined in restyle

rreusser:
This PR adds docs and a codepen explaining the behavior of null vs. undefined. (null affirmatively resets to default, undefined is understood as unspecified and is ignored.) It's a subtle-but-useful feature that was added in https://github.com/plotly/plotly.js/pull/844


cldougl:
nice @rreusser !!
I think it may be helpful to add a `null/null` ("reset") button in so that users can see how each option affects the graph from the beginning state.

rreusser:
Good idea. Updated!

cldougl:
💃 💃

PR #705, opened 2017-02-28
Fixed "Line Shape Options for Interpolation" example

SebastianNiemann:


SebastianNiemann:
My false, I read the contribution guideline to late. I will submit another PR fixing the error in the correct branch.

PR #706, opened 2017-02-28
Fixed "Line Shape Options for Interpolation" example

SebastianNiemann:


cldougl:
thank you @SebastianNiemann !!

PR #709, opened 2017-03-02
changed feb day in gantt chart to realistic value (<=28)

Kully:


Kully:
@cldougl Funny error with the Gantt docs. Turns out I had a lot of `Feb 30th` in there, so the bars disappeared as a result.

Changed to actual feb days.

cldougl:
@Kully nice! thanks for taking a look
💃

PR #723, opened 2017-03-17
added ex for gantt

Kully:


Kully:
@cldougl ready for a review
meant to show `group_tasks=True` used

cldougl:
💃

PR #729, opened 2017-03-22
fixed up share_key example in dashboard doc

Kully:
@cldougl I fixed up the sharekey function vis a vis that one support q


cldougl:
awesome 💃

PR #758, opened 2017-05-10
Add markdown to dboard

Kully:
@cldougl


Kully:
I already updated the image of the dboard in S3

Issue #805, opened 2017-08-28
GitHub corner links should link to documentation page on GitHub


jackparmer:
Currently, the GitHub link goes to the project repository root. It might make more sense at this point to link to the doc page on GitHub instead.

More community members might contribute to the ever-expanding docs if it were easier to go from the doc page to its source on GitHub.

Kully:
@jackparmer I assume you are talking about this:
![screen shot 2017-09-12 at 2 50 27 pm](https://user-images.githubusercontent.com/10369095/30343002-c90ae6aa-97c9-11e7-80a6-96e4b75bc645.png)

I'm trying to understand the goal here. Is it to encourage folks to contribute to docs over the actual source code of charts, etc _or_ is an observation that more people want to edit the documentation and not the chart code?

jackparmer:
yeah, it's to encourage more community members to make updates to the docs themselves

michaelbabyn:
@Kully  I've got a fix for this, do you mind if I take it?

Kully:
@michaelbabyn go for it!

PR #814, opened 2017-09-14
corner github links to documentation page

Kully:
Resolves Issue: https://github.com/plotly/documentation/issues/805

### DIFF
--- a/_includes/breadcrumb.html
+++ b/_includes/breadcrumb.html
@@ -59,7 +59,7 @@
                 <span>Fork on Github</span>
             </a>
             {% elsif page.language == 'python' or page.language == 'pandas' %}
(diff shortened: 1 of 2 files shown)


Kully:
@jackparmer ready for review

jackparmer:
It would be nice if Jekyll had a way to link the actual documentation page and not the root folder for the language. I don't think anyone is going to navigate through our Jekyll folder to fix a type or add a new example.

Kully:
> It would be nice if Jekyll had a way to link the actual documentation page and not the root folder for the language. I don't think anyone is going to navigate through our Jekyll folder to fix a type or add a new example.

Yes, I agree that would be best. I'll see if I can find a way to link them individually like that.

Kully:
closing in favor of https://github.com/plotly/documentation/pull/1030

PR #820, opened 2017-09-27
fixes back link for /javascript

nicolaskruchten:
On some pages like https://plot.ly/javascript/getting-started/ the "Back to Plotly_js" link points to `/plotly_js` instead of `/javascript` so the user lands on https://plot.y/create instead of where they want to go. 

I copied the fix from `_layouts/base.html` into the files where it was missing.


cldougl:
perfect 💃  !

PR #825, opened 2017-10-04
added mri brain doc page

Kully:
@cldougl @jackparmer 
ready for a look

P.S. My comment about the grid upload failure is not entirely justified. I was able to create a grid just now with my own account, but not with the PlotBot account meant for the documentation. Could this be an issue of lots of charts being uploaded/rerun by that account today?

---
This was the error:
```
PlotlyRequestError: This file is too big! Your current subscription is limited to 10,485 KB uploads. For more information, please visit: https://plot.ly/settings/subscription.
```


jackparmer:
💃

cldougl:
@Kully for the python docs you can use `PythonPlotBot` https://plot.ly/~PythonPlotBot 
can you add the notebook to the publisher cell so that users can download it here:
screen shot 2017-10-05 at 9 21 46 am

PR #827, opened 2017-10-06
Dash links

chriddyp:


chriddyp:
adding dash links on several pages. did i miss anywhere? 
also, how do I add new thumbnails? I couldn't find it anywhere in [Contributing.md](https://github.com/plotly/documentation/blob/source-design-merge/Contributing.md), it looks like they're being sourced from images.plot.ly now.

@Kully - could you test these changes?

cc @cldougl @bcdunbar

chriddyp:
@cldougl - I noticed that there is some new symmetry in the docs now: 5 thumbnails per row. we have room for this one in the "Add Custom Controls" row but not in the top "Plotly Fundamentals" row. Anything in there that we can kick out to make room for Dash?

cldougl:
maybe the jupyter notebook tutorial - imo the static img and updating ex should stay there since we were asked so often about those subjects

cldougl:
Instead of adding it to Plotly Fundamentals- perhaps we should make a Dashboards and Reports row? (Dash / Dashboard API / some of these: https://plot.ly/python/#report-generation w/ thumbnails)

Kully:
> @Kully - could you test these changes?

Yes!

Kully:
@chriddyp I like this thumbnail more than the one you had set. I think it's too similar to its neighbour and doesn't pop out enough:

dash_thumbnail

In context:
screen shot 2017-10-06 at 3 03 52 pm

Kully:
@chriddyp permalinks are fixed and it's loading properly. I am currently in the process of hiding the Jupyter Notebook thumbnail tutorial

Kully:
After deciding on the choice of thumbnail, I give it a 💃

chriddyp:
great, thanks @Kully ! that thumbnail looks good 👍  where are the thumbnails stored? could you also make sure that the image size is small?

> Instead of adding it to Plotly Fundamentals- perhaps we should make a Dashboards and Reports row? (Dash / Dashboard API / some of these: https://plot.ly/python/#report-generation w/ thumbnails)

yeah, that's a great idea. "Dashboards, Apps, and Reports" could definitely make it's own row. @cldougl maybe we can merge this for now and then add that section next?

Kully:
> where are the thumbnails stored?

Number 10: https://github.com/plotly/documentation/tree/source-design-merge/_posts/python

Kully:
> could you also make sure that the image size is small

Yes, I've already made the image 160by160 pxls, the standard size for thumbnails

chriddyp:
thanks for reviewing @Kully ! could you deploy this when you get a chance?

Kully:
> thanks for reviewing @Kully ! could you deploy this when you get a chance?

sure. I'll merge it now and deploy

PR #829, opened 2017-10-10
fixed dash links in plotly/python

Kully:
@chriddyp @cldougl 
Resolves: https://github.com/plotly/documentation/issues/828

### DIFF
--- a/_posts/python/controls/2017-10-05-dash-controls.html
+++ b/_posts/python/controls/2017-10-05-dash-controls.html
@@ -1,5 +1,5 @@
 ---
-permalink: /plot.ly/products/dash/
+permalink: //plot.ly/products/dash/
 name: Dash - Controls with Python Callbacks (NEW!)
 thumbnail: thumbnail/dash_thumbnail_160.jpg
 language: python
--- a/_posts/python/controls/2017-10-05-dash-controls.html
+++ b/_posts/python/controls/2017-10-05-dash-controls.html
@@ -1,5 +1,5 @@
 ---
-permalink: //plot.ly/products/dash/
+permalink: /plot.ly/products/dash/
 name: Dash - Controls with Python Callbacks (NEW!)
 thumbnail: thumbnail/dash_thumbnail_160.jpg
 language: python
--- a/_posts/python/controls/2017-10-05-dash-controls.html
+++ b/_posts/python/controls/2017-10-05-dash-controls.html
@@ -1,5 +1,5 @@
 ---
-permalink: /products/dash/
+permalink: /plot.ly/products/dash/
 name: Dash - Controls with Python Callbacks (NEW!)
 thumbnail: thumbnail/dash_thumbnail_160.jpg
 language: python
--- a/_posts/python/fundamentals/2017-10-05-dash.html
+++ b/_posts/python/fundamentals/2017-10-05-dash.html
@@ -1,5 +1,5 @@
 ---
-permalink: /plot.ly/products/dash/
+permalink: //plot.ly/products/dash/
 name: Dash - Interactive Python Apps (NEW!)
 thumbnail: thumbnail/dash_thumbnail_160.jpg
 language: python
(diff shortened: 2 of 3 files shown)


_posts/python/fundamentals/2017-10-05-dash.html
```diff
@@ -1,5 +1,5 @@
 ---
-permalink: /products/dash/
+permalink: /plot.ly/products/dash/
```
chriddyp: In why doesn't https://plot.ly/products/dash work? In https://github.com/plotly/documentation/pull/827/commits/07c76f6502087c9c1ee3c54f639d4fb1d1d5dd16 you changed it from `https://plot.ly/products/dash` to `/products/dash` but in `documentation_eg.html` the `href` is the `page.permalink`: https://github.com/plotly/documentation/blob/source-design-merge/_includes/documentation_eg.html#L157

Kully: `https://plot.ly/products/dash` was not working in the original PR but my first change to `/products/dash` also didn't work. It needed to be `/plot.ly/products/dash`.

Kully: ^Regarding your second question, I'm not sure. Just based off of serving it locally, it goes to the correct link after this change above

chriddyp: That's odd. This feels like a hack - `/plot.ly/products/dash/` is only working because this line:
```
                <a href="{% if page.permalink contains 'http' %}{{page.permalink}}{% else %}/{{page.permalink}}{% endif %}">
```

inserts a leading `/` into the url so that we have `<a href="//plot.ly/products/dash">` and a leading in a `//` is valid (see https://stackoverflow.com/questions/9646407/two-forward-slashes-in-a-url-src-href-attribute).

I don't understand why `{% if page.permalink contains 'http' %}` isn't `true` in this scenario. Could you try it again with `https://plot.ly/products/dash` and manually check the `<a/>` tag in the rendered HTML to see what it looks like?

Kully: The problem is I'm getting this ERROR:
```
ERROR: YOUR SITE COULD NOT BE BUILT:
                    ------------------------------------
                    The URL /https:/plot.ly/products/dash/ is invalid because it contains a colon.
```
FYI there's a github dead issue about it [here](https://github.com/jekyll/jekyll/issues/5284#issuecomment-291815611).

Kully:
@chriddyp I've add `//` to the beginnings of the permalinks

_posts/python/fundamentals/2017-10-05-dash.html
chriddyp: ah ok, that makes sense. I guess `{% if page.permalink contains 'http'` never worked then (or it only worked in old versions).

What if instead we used `permalink: //plot.ly/dash/products` but we changed the logic to be:
```
{% if (page.permalink | slice: 0) == '//' %}
    {{page.permalink}}
{% else %}
    /{{page.permalink}}
{% endif %}
```

If we _only_ change permalink to `//` without changing the logic then the link will end up being `///plot.ly/dash/products` which may work but isn't semanatically correct. This logic will keep it as `//plot.ly/dash/products` which is semantic.

Kully: I like this idea. I'll look at the surrounding code to see if this won't break anything then write it in

Kully:
@chriddyp looks good to you?

_posts/python/controls/2017-10-05-dash-controls.html
```diff
@@ -1,5 +1,5 @@
 ---
-permalink: /products/dash/
+permalink: /plot.ly/products/dash/
```
chriddyp: missing `//` here?

chriddyp:
💃  once my last comment is addresssed

PR #863, opened 2017-11-23
Bullet Chart Doc

Kully:
@cldougl waiting to merge bullet-charts in plotly.py repo


_posts/python/statistical/distplot/2015-06-30-distplots.html
```diff
@@ -10,7 +10,7 @@
 title: Python Distplots | plotly
 display_as: statistical
 has_thumbnail: true
-page_type: example_index
+page_type: u-guide
```
cldougl: Can you leave the distplots on the index page ?

Otherwise it looks like:

screen shot 2017-11-24 at 3 25 11 pm

_posts/python/statistical/bullet/2015-06-30-bullet-chart.html
```diff
@@ -0,0 +1,508 @@
...
+name: Bullet Charts
+has_thumbnail: true
+thumbnail: thumbnail/bullet_charts.jpg
+layout: user-guide
+language: python
+title: Python Bullet Charts | plotly
+display_as: statistical
+has_thumbnail: true
+page_type: u-guide
+order: 0
```
cldougl: typically we go from more basic -> more advanced so this should be ordered further down

_posts/python/statistical/bullet/2015-06-30-bullet-chart.html
```diff
@@ -0,0 +1,508 @@
...
+
+</div>
+</div>
+
+</div>
+<div class="cell border-box-sizing text_cell rendered"><div class="prompt input_prompt">
+</div>
+<div class="inner_cell">
+<div class="text_cell_render border-box-sizing rendered_html">
+<h4 id="Simple-Bullet-Chart">Simple Bullet Chart<a class="anchor-link" href="#Simple-Bullet-Chart">&#194;&#182;</a></h4><p>Stephen Few's Bullet Chart was invented to replace dashboard gagues and meters, combining both types of charts into simple bar charts with qualitative bars (<code>ranges</code>), quantitiative bars (<code>measures</code>) and performance points (<code>markers</code>) all into one simple layout. <code>ranges</code> typically are broken into three values: bad, okay, good, the <code>measures</code> are the darker bars that represent the actual values that a particular variable reached, and the points or <code>markers</code> usually indicate a goal point relative to the value achieved by the measure bar.</p>
```
cldougl: I think there's a typo, should be gauges?

_posts/python/statistical/bullet/2015-06-30-bullet-chart.html
```diff
@@ -0,0 +1,508 @@
...
+
+</div>
+</div>
+
+</div>
+<div class="cell border-box-sizing text_cell rendered"><div class="prompt input_prompt">
+</div>
+<div class="inner_cell">
+<div class="text_cell_render border-box-sizing rendered_html">
+<h4 id="Custom-Kwargs">Custom Kwargs<a class="anchor-link" href="#Custom-Kwargs">&#194;&#182;</a></h4>
```
cldougl: for this I would probably edit the legend visibility so it's not so repetitive

_posts/python/statistical/bullet/2015-06-30-bullet-chart.html
Kully: I put it on the front because it's new to the Plotly Universe. Are we not concerned with advertising that in some way?

cldougl: I'm not super concerned. If you feel like it needs to be advertised, a better way would be to put it on the index page, but the order (on the index page as well as the stats page) should stay basic -> advanced.

_posts/python/statistical/distplot/2015-06-30-distplots.html
Kully: sure

I vote for putting Bullet before Violin

_posts/python/statistical/bullet/2015-06-30-bullet-chart.html
Kully: if I do something like just hiding the repetitive symbols, then users won't have a way to deselect parts of the chart. I would vote against this as it's a custom option anyways.

Perhaps I can not put it in the example doc

cldougl: One thing you could do is group the components of each bullet and then hide the remaining components. I think having the option to deselect the entire bullet is a more common usecase than deselecting individual elements of the chart (usually when this type of chart is created the entire bullet is a component rather than seeing all of the individual components that make up the bullet in legend)

Kully: good idea!

Kully: but that will require me to patch up the bulet function correct? It's already released fyi

cldougl: you can do so by editing the figure after if you prefer- I think that's ok!

Kully:
@cldougl all comments addressed :)

cldougl:
perfect!
is it the correct version in the doc now (in the version check section)
if that's all set then it's good to 💃  !

PR #864, opened 2017-11-28
fixed first subplot table

Kully:
@cldougl 
not sure why but the first subplot of tables in https://plot.ly/python/table-subplots/ still had an incorrect layout vis a vis Jack's comment in `python-api` in Slack

I corrected the format now. Must have by accident deleted the code and used the incorrect format of that subplot before the old PR merged.

### DIFF
--- a/_posts/python/subplots/table-subplots/2015-06-30-table-subplots.html
+++ b/_posts/python/subplots/table-subplots/2015-06-30-table-subplots.html
@@ -28,7 +28,7 @@ Run  <code>pip install plotly --upgrade</code> to update your Plotly version</p>
 </div>
 <div class="cell border-box-sizing code_cell rendered">
 <div class="input">
-<div class="prompt input_prompt">In&nbsp;[1]:</div>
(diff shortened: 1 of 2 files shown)


cldougl:
just 2 things I noticed:
1. the headers in the first table
screen shot 2017-11-29 at 12 09 27 pm
2. the last table got messed up again (looks like the dataframe isn't getting referenced correctly)
screen shot 2017-11-29 at 12 09 06 pm

let me know if you need help w/ either of those! I think it might help to reduce the left margin as well

cldougl:
the first table still has a few spacing issues (looks like the dates are getting cut off and some of the headers are still getting cut off)

these changes are a good improvement though so it's okay if you want to merge this and then make a new pr for the spacing of the first table 👍  💃

Kully:
> these changes are a good improvement though so it's okay if you want to merge this and then make a new pr for the spacing of the first table 👍 💃

I'd rather tackle all issues in this one PR

cldougl:
💃

Issue #881, opened 2018-02-21
florida county choropleth doc page is broken


Kully:
URL: https://plot.ly/python/mapbox-county-choropleth/

First issue arrises with this line:
`florida_data = ast.literal_eval(florida_data)`

cldougl:
@Kully was a cell removed from this notebook?

![screen shot 2018-03-08 at 1 55 39 pm](https://user-images.githubusercontent.com/2789078/37170304-7c66162c-22d8-11e8-86ae-c97d67e0483b.png)

Kully:
> was a cell removed from this notebook?

I don't remember if I accidentally deleted one, but I do not think so.

PR #890, opened 2018-03-08
fix florida county doc page (mapbox in maps)

Kully:
fixes: https://github.com/plotly/documentation/issues/881


Kully:
@cldougl only issue for me is a warning message that appears after one of the cells:
```
UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
```

Does you want this removed?

cldougl:
that's ok- thanks for fixing this! 💃

PR #896, opened 2018-03-13
documenting Plotly.react

### LINKED ISSUE #886: Plotly.react
`Plotly.react` should be described in https://plot.ly/javascript/plotlyjs-function-reference/

Key points:
- signature matches `Plotly.newPlot`
- describe [datarevision](https://github.com/plotly/plotly.js/blob/113b3618732f28d1a2158771544bf4e60e00c084/src/plots/layout_attributes.js#L186)
- same end results as `newPlot` with the speed of `restyle`/`relayout`

nicolaskruchten:
Fixes #886


cldougl:
💃 now that the serving mystery 🔍  was solved!

PR #897, opened 2018-03-13
smooth scroll hash offset

### LINKED ISSUE #892: attribute hash links are scrolled off-screen
For example, I wanted to point someone to the reference to parcoords.dimensions.tickformat at https://plot.ly/javascript/reference/#parcoords-dimensions-tickformat, but it's not visible when you visit that link - the top item you see is the "visible" attribute, which is next after tickformat. I guess it's scrolled under the floating header...
screen shot 2018-03-11 at 7 40 35 pm

nicolaskruchten:
Resolves #892 via javascript


cldougl:
💃 !

PR #923, opened 2018-04-24
Spelling and markdown style fixes

MercifulCode:


cldougl:
@MercifulCode thanks for these edits!! @Kully would you be able to fix the conflicts here?

PR #931, opened 2018-05-22
Splom Doc

Kully:
This is the thumbnail:
splom_image

It appears after scatterplotmatrix doc, so not on the front Plotly page.

cc. @jackparmer


jackparmer:
Looks good! Let's publish it at plot.ly/python/splom 💃 

@bcdunbar Please do the same thing at plot.ly/r/splom ⚖️

PR #965, opened 2018-07-04
exclude 'info_array' attrs from extra-info-about "items" block

etpinard:
Currently, 'info_array' attributes have an erroneous _Each object has one or more of the keys listed below_ line e.g. for axis ranges:

image

with this patch, we now have:

image

cc @cldougl @nicolaskruchten


nicolaskruchten:
💃 if it looks good locally, i haven't built it myself

Issue #969, opened 2018-07-17
Bad example on website (exponentformat "All" instead of "all") 


cldougl:
- [x] In the [Set and Style Axes Title Labels and Ticks](https://plot.ly/javascript/axes/#set-and-style-axes-title-labels-and-ticks) example, `layout.yaxis.showexponent` is set to 'All', but actually this parameter is case-sensitive and should be "all". See [live example here](https://jsbin.com/karasol/4/edit?js,output).

- [x] Also update dataset so that it is large or small enough to actually use exponents

PR #970, opened 2018-07-17
fix bug in showexponent

michaelbabyn:
fixed case in 'All' to 'all' for showexponent attribute. from [this issue](https://github.com/plotly/documentation/issues/969) @cldougl @bcdunbar


cldougl:
great! 💃

Issue #979, opened 2018-07-25
Bundler needed to be updated - "bundle install" issue


priyatharsan:
Failed installation of Gemfile due to the dependency mismatch after installing Ruby 2.4.2 (current ruby version according to https://pages.github.com/versions/) 
![bundle_install_issue](https://user-images.githubusercontent.com/25846124/43226791-ceea79aa-902a-11e8-8354-fb28ea92643a.PNG)

priyatharsan:
Closing this issue as the `Gemfile` is now been updated.

PR #980, opened 2018-07-26
update nokogiri x64 gem

bcdunbar:
Resolves https://github.com/plotly/documentation/issues/979

if `jekyll serve --config _config_dev.yml` still doesn't work, try `bundle exec jekyll serve --config _config_dev.yml`

cc @cldougl @priyatharsan


cldougl:
thanks @bcdunbar !!

Issue #937, opened 2018-05-31
Add info about new legend events 


etpinard:
from https://github.com/plotly/plotly.js/pull/2581#issuecomment-393603619

jacobq:
In particular, I'd like to see examples / recommendations regarding how to prevent firing the click handlers in the case of doubleclick events. (IIRC, this problem also occurs when dealing directly with the DOM, e.g. https://output.jsbin.com/nikimir/3)

michaelbabyn:
closed by [#1002](https://github.com/plotly/documentation/pull/1002)

PR #1002, opened 2018-08-06
Legend click event

michaelbabyn:
I updated plot.ly/javascript/plotlyjs_events with a new section on legend click events to deal with [this issue](https://github.com/plotly/documentation/issues/937). 

Unfortunately, the only way I can access this page is through the link or by searching at the moment.
@cldougl  , Is this what we want or would it be better for a link to this page to appear here:

javascript chart events


cldougl:
@michaelbabyn the event doc should be linked in the sidebar:

screen shot 2018-08-07 at 11 16 29 am

If you think it would be more discoverable to add a thumbnail link to https://plot.ly/javascript/#chart-events you're more than welcome :) I think that sounds like a good idea!

Issue #967, opened 2018-07-11
HTML isn't getting escaped in this example


chriddyp:
so we see the bold and line breaks instead of `<b>` and `<br/>`
![image](https://user-images.githubusercontent.com/1280389/42598524-72dd99f4-852a-11e8-9417-8af03fc3b606.png)

PR #1006, opened 2018-08-07
style annotation html escape

Kully:
https://github.com/plotly/documentation/issues/967#event-1775628706

Work In Progress
<i> needs to become `&lt; i &gt;`


jdamiba:
Change looks good to me. 💃

Kully:
> Change looks good to me. 💃

It's not ready yet.

michaelbabyn:
@Kully , I think this issue would be fixed if we passed the example content through Jekyll's `xml_escape` filter. We can do that by changing [this line](https://github.com/plotly/documentation/blob/source-design-merge/_includes/auto_examples.html#L15) of code to `<pre><code class="{{page.language}}">{{ example.content | xml_escape }}</code></pre>`

Jekyll takes care of doing the escaping and I'm not aware of any examples where we use html within the code, what do you think?

Kully:
> I'm not aware of any examples where we use html within the code

Isn't example from the issue one? :https://plot.ly/javascript/text-and-annotations/#styling-and-formatting-annotations

michaelbabyn:
Sorry, I was unclear. In that case we have html in the codeblock but it isn't being escaped. The solution I'm proposing will escape all html in all code blocks so if we are using the html to style any code in these codeblocks it would be escaped by that solution. So when I said I'm not aware of any example when we use html, I meant I'm not aware of any example where escaping the html would lead to unintended consequences.

Kully:
Okay, thank you for the clarification. I tried it and unfortunately it wasn't showing the change when I jekyll served it locally.

I may have done some tinkering with the actual HTML blob, moved <br>s inside of strings or whatnot, so I'll make sure all of that is in order.

Issue #938, opened 2018-05-31
colorway default is stringified


alexcjohnson:
![screen shot 2018-05-31 at 1 55 53 pm](https://user-images.githubusercontent.com/2678795/40799017-7c8f46ca-64da-11e8-94ca-16f0faa2afb0.png)

Should be formatted as the array it is:
```js
["#1f77b4", "#ff7f0e", "#2ca02c", "#d62728", "#9467bd", "#8c564b", "#e377c2", "#7f7f7f", "#bcbd22", "#17becf"]
```

PR #1014, opened 2018-08-09
Colorway correction in reference docs

Kully:
Fixes https://github.com/plotly/documentation/issues/938


cldougl:
@Kully will this get overwritten if we update `_data/plotschema.json` programatically (ie during deploy)? If so, do you think you could come up w/ a more permanent solution?

Kully:
Now each deploy is correcting the format of colorway so that appears as a list of colors in the reference page. The solution is to add `\'`s around the brackets of the colorway list:

`"dflt": '["#1f77b4","#ff7f0e",...]'`

Right now, the script _only_ looks for a `dflt`that points to a 10-item list each which start with `#`

Kully:
> can you resolve the conflicts on this branch

sure thing

Kully:
@cldougl resolved conflicts

Issue #856, opened 2017-11-03
copy to clipboard feature is broken


cldougl:
![screen shot 2017-11-03 at 5 28 23 pm](https://user-images.githubusercontent.com/2789078/32396646-794a4b3a-c0bc-11e7-8a7b-692eb9360f2a.png)

PR #1021, opened 2018-08-10
Copy clipboard Box

Kully:
For Issue https://github.com/plotly/documentation/issues/856


jdamiba:
I was able to successfully build this branch on my local machine and saw that the "Copy to clipboard!" is indeed in visible. However, on my machine the box is just text, and not clickable to add the text to my clipboard. Is that related to how the assets are loaded in a localhost environment vs production?

Kully:
Report of the Issue so far:

Google (Chrome) has been moving away from flash over the years to the point that you need to allow flash to play on most websites.

The copyCode buttons rely on flash so this means that we need to allow each time, unless the button has a bigger resolution that 400X300 (not gonna happen).

If I do allow flash to run on the page, I am still hitting an error that I am working to fix now, but I am wondering if we should be moving away from flash altogether? I can't imagine anyone who would want to enable flash on Chrome before simply clicking a simple `Copy` button.

Thoughts?

Kully:
relevant issue: https://github.com/zeroclipboard/zeroclipboard/issues/589

Kully:
clipboard-works

Kully:
@cldougl ready for a review - clipboard works now.

NOTE: I've only tested in Chrome. I'll test in Safari as well to be very sure.

cldougl:
@Kully can you use browserstack and test all browsers?

Kully:
Testing browsers: (OS: Mojave)
- [X] Firefox
- [ ] Opera
- [ ] Chrome
- [ ] Safari
- [ ]

nicolaskruchten:
@michaelbabyn can you see if this works/can be merged please? copy-to-clipboard right now in prod does NOT work in Chrome/OSX for me

michaelbabyn:
This works for me in chrome and firefox although on Chrome it's not clear without opening dev tools that you need flash set up and enabled for the button to work.

Issue #954, opened 2018-06-13
not referencing files


Kully:
The example https://plot.ly/python/ternary-scatter-contour/#load-and-process-data-files is broken. Files not found: should be referencing data files in plotly/datasets repo

bcdunbar:
Had a quick look but didn't come across these data files anywhere. 

@Kully @cldougl Do either of you know where to find these 2 JSON files.

Kully:
finished

PR #1028, opened 2018-08-16
use scatter and contour files from datasets

Kully:
fixes https://github.com/plotly/documentation/issues/954

### DIFF
--- a/_posts/python/scientific/ternary-scatter-contour/2015-06-30-ternary-scatter-contour.html
+++ b/_posts/python/scientific/ternary-scatter-contour/2015-06-30-ternary-scatter-contour.html
@@ -14,24 +14,22 @@ order: 11
 page_type: u-guide
 ---
 {% raw %}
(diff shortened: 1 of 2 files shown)


Kully:
@bcdunbar issue is solved - the missing JSON files are now on the web so they can be referenced and read

bcdunbar:
@Kully thanks!

Looks like the plots may be private here.

Kully:
> Looks like the plots may be private here.

> Looks like the plots may be private here.

The plots are not private, not live or checking with `jekyll serve`. The url links point to the right json files too. What are you noticing specifically that is private?

Kully:
> Oh I meant when I open https://plot.ly/~jordanpeterson/1125.embed (and the other plots) I get a 404

Okay, I made this and other relevant plots public. Can you verify it's not private on your end?

jdamiba:
I can verify that  https://plot.ly/~jordanpeterson/1125.embed is not private on my end. 💃

PR #1031, opened 2018-08-17
An updated version of widget_app using the FigureWidget

kmader:
Changing the widget_app to be self-contained (making a figure from scratch rather than reusing the one on the plotly gallery) and use the new FigureWIdget approach along with the `.batch_update()` style of updating


jackparmer:
@michaelbabyn or @Kully - please review and merge

michaelbabyn:
@kmader, Thank you for this! I will merge this and then update the metadata so we can publish this to our documentation page.

PR #1032, opened 2018-08-17
Update python layout images doc

michaelbabyn:
I added an example showing how to zoom on a background image without distorting it and got rid of the deprecation warnings.


priyatharsan:
Great job on the doc. Looks good, except in the "Facial Recognition Overlay" example, the image is not zooming along with the rectangles.

bcdunbar:
This looks great. As far as this PR goes 💃  

That said, some suggestions to make this doc 👌 (not necessarily apart of this PR but probably easy to add on to this)

- organize and spread out imports (https://github.com/plotly/documentation/pull/1048#discussion_r212608405), like:

```
import plotly.plotly as py
import plotly.tools as tls

import matplotlib.pyplot as plt
import numpy as np
```

- The **Interactive Facial Recognition Overlays** is great but it may be easier to read the example if the code is laid out like the previous examples.

PR #1045, opened 2018-08-21
replace developer suport and workshops links with Dash link in header

Kully:
resolves https://github.com/plotly/streambed/issues/11339

PR #1056, opened 2018-08-27
Added navbar change to mobile

jdamiba:
follow up to https://github.com/plotly/documentation/pull/1045

making the change in the mobile nav menu as well.


Kully:
My guess is that the arrow is part of the `Show Sidebar` icon (right arrow a line above and below that are parallel).

Issue #1064, opened 2018-09-01
Full Reference search links fail for array properties:


jonmmease:
#### To reproduce:
 - Visit https://plot.ly/python/reference/
 - Search for "shape layer"
 - Click the top result "Layout > shapes > items > shape > layer". Nothing happens

#### What's going wrong:
The URL that this search result points to is https://plot.ly/python/reference/#layout-shapes-items-shape-layer.

But if you scroll down to find the `layout.shape.layer` property manually, it has url 
https://plot.ly/python/reference/#layout-shapes-layer (without "item")

Issue #925, opened 2018-04-25
clicking searched items within the plotly.js docs does not jump to the item 


SKalt:
Steps to reproduce this issue:
1. go to //plot.ly/javascript/reference/
2. type 'button' into the search bar
3. click any of the search items

At this point if you clicked the first item, the URL is https://plot.ly/javascript/reference/#layout-xaxis-rangeselector-buttons-items.  `document.getElementById('layout-xaxis-rangeselector-buttons-items')` returns an empty NodeList. Manually navigating to layout > xaxis > rangeselector > buttons shows no subitems labeled "items". Instead, there are other subelements corresponding to each of the properties of  a layout >... > buttons > item.

It looks like this issue is the search going out-of-sync with what's actually on the page.

michaelbabyn:
closed by #1069

PR #1069, opened 2018-09-04
improve search link for array properties

michaelbabyn:
Fixes https://github.com/plotly/documentation/issues/1064


michaelbabyn:
I noticed that any nested attributes deeper than 4 or 5 levels didn't get added to the angolia search so I also updated the ref_search script so that it doesn't max out. Now this PR also fixes https://github.com/plotly/documentation/issues/925.

PR #1079, opened 2018-09-07
Add Dash Ads Inline

jdamiba:
The ad is showing on the Python page but not the JavaScript or R page.


_includes/documentation_eg.html
```diff
@@ -717,7 +721,7 @@ <h4 class="example-section-title" id="layout-options">
 {% if databases %}
 <section class="--tutorial-section no-image" id="databases">
     <header class="--section-header"><a href="#databases">Connecting to Databases</a>
-        <div class="icon copy" data-tooltip="Click to copy direct link.">
+        <div class="icon copy" data-tooltip="Click to copy direct lin     k.">
```
cldougl: I think this may be a typo here

_includes/documentation_eg.html
```diff
@@ -627,6 +627,10 @@ <h4 class="example-section-title" id="layout-options">
 </div>
 {% endif %}
 
+{% if page.language == "python" or page.language == "r" or page.language == "javascript" %}
```
cldougl: this is working for me on /python and /r
for https://plot.ly/javascript the language is actually "plotly_js"
for example: https://github.com/plotly/documentation/pull/1079/files#diff-3c5b21462f63fc69d620c11747257935R501

jdamiba:
I think this should be ready for review now. The ad is now visible on all 3 pages.

cldougl:
💃 lgtm now! 
Let's see if this creates more traffic and then we can assess if we want to change the placement or make any further edits from there!

Issue #1065, opened 2018-09-03
plotly.animate missing from plotly.js function reference


chriddyp:
That is, from this page: https://plot.ly/javascript/plotlyjs-function-reference/

michaelbabyn:
closed by #1084

PR #1084, opened 2018-09-11
add Plotly.animate to js function reference.

michaelbabyn:
fixes https://github.com/plotly/documentation/issues/1065


_posts/reference_pages/2016-06-03-plotly_js_function_ref.html
```diff
@@ -443,6 +443,57 @@ <h4 id="plotly-prependtraces"><a href="{{ BASE_URL }}/javascript/plotlyjs-functi
...
 </code></pre>
 
+<h4 id="plotly-animate"><a href="{{ BASE_URL }}/javascript/plotlyjs-function-reference/#plotlyanimate">Plotly.animate</a></h4>
+Add dynamic behaviour to plotly graphs with <code>Plotly.animate</code>.
+
+<br /><br />
+<fieldset class="signatures">
+<legend>Signature</legend>
+<dl>
+  <dt><code>Plotly.relayout(graphDiv, frame(s), animationAttributes)</code></dt>
```
nicolaskruchten: this should say `Plotly.animate(...` right?

_posts/reference_pages/2016-06-03-plotly_js_function_ref.html
```diff
@@ -443,6 +443,57 @@ <h4 id="plotly-prependtraces"><a href="{{ BASE_URL }}/javascript/plotlyjs-functi
...
+      <dd>An object, see <a href="{{ BASE_URL }}/javascript/animations">documentation</a> for examples.</dd>
+    </dl>
+  </dd>
+</dl>
+</fieldset>
+<br /><br />
+
+
+<pre><code class="language-javascript hljs" data-lang="javascript">
+Plotly.plot('graph', [{
```
nicolaskruchten: this should use `newPlot` rather than `plot` as the latter is deprecated

_posts/reference_pages/2016-06-03-plotly_js_function_ref.html
```diff
@@ -443,6 +443,57 @@ <h4 id="plotly-prependtraces"><a href="{{ BASE_URL }}/javascript/plotlyjs-functi
...
+<br /><br />
+<fieldset class="signatures">
+<legend>Signature</legend>
+<dl>
+  <dt><code>Plotly.animate(graphDiv, frame(s), animationAttributes)</code></dt>
+  <dd>
+    <dl>
+      <dt><code>graphDiv</code></dt>
+      <dd>DOM node or string id of a DOM node</dd>
+      <dt><code>frame(s)</code></dt>
```
nicolaskruchten: Could we provide a succinct explanation of just what a "frame" is intended to be in this context? Either here or at the top of the page where the "common parameters" are laid out.

@etpinard @alexcjohnson  from our conversation last Thursday the idea I came away with is that it's got the same structure as a full figure (minus the `frames` key!) but is intended to only contain the deltas? Like if I have a figure with a bunch of traces and layout information and I want to animate the title color, the "frame" would contain only `{layout: {title: {something}}`... is that right?

nicolaskruchten: From reading the examples, it looks like this argument can also be a list of strings, and that frames can also have a `name` key?

etpinard: > and I want to animate the title color, the "frame" would contain only `{layout: {title: {something}}`... is that right?

that's correct.



> and that frames can also have a `name` key?

that's correct. You can point to frames "already added to the figure" via `Plotly.addFrames`  or `Plotly.newPlot(gd, {data: [], layout: {}, frames: [/* */]})` by their name.

etpinard: More info: 

https://github.com/plotly/plotly.js/blob/344a6c560fd046657b4e662a046f6141aaf2df29/src/plot_api/plot_api.js#L2601-L2627

nicolaskruchten: Thanks @etpinard ! What's a "group"? Do frames take a `group` key as well as a `name` key?

etpinard: > Do frames take a `group` key as well as a `name` key?

yeah, they _can_ take a group -> https://github.com/plotly/plotly.js/blob/5fa493aca6d820580aca6986fb505d4aad729911/src/plots/frame_attributes.js#L14-L21

nicolaskruchten:
Nice one @michaelbabyn, this is a really compact yet complete way of documenting these functions :)

💃 as far as I'm concerned!

PR #1088, opened 2018-09-13
Add responsive config opt

antoinerg:
Updated doc to reflect the new `responsive` config option available since plotly.js v1.41.0!


antoinerg:
cc @cldougl and @bcdunbar

Issue #1018, opened 2018-08-10
joy plot/ridge plot examples


chriddyp:
Someone in the community forum asked about Ridgeline and Joy Plot charts in Python: https://community.plot.ly/t/ridgeline-joy-plots-in-dash-automatic/12602/3

this might be a nice page of content. https://plot.ly/python/joy-plots. The plots are beautiful too :)

chriddyp:
> Your second link is inactive. Could you please repost?

Sorry this was unclear. This link  https://plot.ly/python/joy-plots is just meant to be the URL of where these new examples should go. It doesn't exist yet.

chriddyp:
Some nice conversation continues in https://community.plot.ly/t/ridgeline-joy-plots-in-dash-automatic/12602/8. In particular, here's a nice example: https://plot.ly/~empet/14632/plotly-joyplotridgelines/#/

chriddyp:
A really nice example here: https://morphocode.com/location-time-urban-data-visualization/
![image](https://user-images.githubusercontent.com/1280389/50254772-646bcf00-03bd-11e9-964e-82ea2711cd53.png)

Kully:
> A really nice example here: https://morphocode.com/location-time-urban-data-visualization/

looks awesome

Mahdis-z:
https://plot.ly/python/violin/

tpvasconcelos:
I wanted to have a little more flexibility into how ridgeline plots are displayed with Plotly (e.g., ensure the correct z-order is used, more flexibility into coloring options, multiple traces per rows, etc...) and ended up building a little fun side project just for this use-case. I'll plug it here in case it can be of help to anyone who runs across this issue: https://github.com/tpvasconcelos/ridgeplot

PR #1090, opened 2018-09-14
Joyplot Doc

Kully:
for Issue: https://github.com/plotly/documentation/issues/1018


Kully:
Not having luck with violin plots...I am going to try building from scratch. Could be a fun FF.

nicolaskruchten:
what's not working with violin plots? Groups as 'y', orientation='h', make them half...

nicolaskruchten:
image

Kully:
> Groups as 'y', orientation='h', make them half...

I'm having trouble allowing them to overlap. `violingap` doesn't do the trick.

nicolaskruchten:
Right, they won't overlap... We could just generalize violin to allow it, rather than wrap a whole subplot-based infrastructure onto it... @etpinard ?

Kully:
> We could just generalize violin to allow it
I'd like to take a first stab at adding that feature if we decide to go that route

Kully:
Continuing off @nicolaskruchten 's point, the violins will not be able to overlap because `violingap` is only allowed in the closed interval $[0, 1]$.

---
If we place violins at y=0, 0.2, and 0.6

newplot

Nudging the 0.6 violin very close to 0.2 (0.21), we get:

newplot 1

---

If we allowed negative numbers for `violingap`, then this would solve the problem.

CODE FOR EXAMPLE:
```javascript
var data = [
{
  type: 'violin',
  x: [0, 5, 7, 8],
  points: 'none',
  box: {
    visible: false
  },
  boxpoints: false,
  line: {
    color: 'black'
  },
  fillcolor: '#8dd3c7',
  opacity: 0.6,
  meanline: {
    visible: false
  },
  y0: 0
},
{
  type: 'violin',
  x: [0, 4, 7, 8],
  points: 'none',
  box: {
    visible: false
  },
  boxpoints: false,
  line: {
    color: 'black'
  },
  fillcolor: '#d3c78d',
  opacity: 0.6,
  meanline: {
    visible: false
  },
  y0: 0.2
},
{
  type: 'violin',
  x: [0, 4, 7, 8],
  points: 'none',
  box: {
    visible: false
  },
  boxpoints: false,
  line: {
    color: 'black'
  },
  fillcolor: '#c78dd3',
  opacity: 0.6,
  meanline: {
    visible: false
  },
  y0: 0.21
}
];

var layout = {
  title: "Basic Horizontal Violin Plot",
  xaxis: {
    zeroline: false
  },
  violingap: 0
};

Plotly.newPlot('graph', data, layout);
```

@etpinard  WDYT?

Kully:
> Ok, and what's wrong again with something like

I should have made them half violins to demonstrate the point. In the example you shared, I want to take the top blue one for example and shift it down a few pixels so that its middle line is somewhere between 1 and 2, while leaving the rest looking the same.

etpinard:
> or example and shift it down a few pixels so that its middle line is somewhere between 1 and 2, while leaving the rest looking the same.

tweaking the `y` coordinates should do the trick, no?

Kully:
> tweaking the `y` coordinates should do the trick, no?

I swapped all the y values of orange violin at 2 to 1.8 and get this: https://codepen.io/plotly/pen/yRXjPe

Can you see how the violins are now all thinner and I was not able to just shove that one violin down by a certain number of pixels?

etpinard:
Thanks for codePen @Kully !

Would something like https://codepen.io/etpinard/pen/yRogaq be ok?

If not, we could add a `width` attribute for violin (and box) traces that would essentially override the internal [`dPos`](https://github.com/plotly/plotly.js/blob/d34015066c8d339f041c9ac5ae68762a342d5d86/src/traces/box/cross_trace_calc.js#L103-L104) parameter. This new attribute would be similar to `width` in bar traces.

Allowing `violingap` and `violingroupgap` to be negative would _work_, as these attributes [affect the violins' width](https://github.com/plotly/plotly.js/blob/d34015066c8d339f041c9ac5ae68762a342d5d86/src/traces/violin/plot.js#L46), but I don't see how we can justify making a "gap" value negative. The "real" parameter we want to override here is the violin width, not the gaps between them.

Kully:
> f not, we could add a `width` attribute for violin (and box) traces that would essentially override the internal [`dPos`](https://github.com/plotly/plotly.js/blob/d34015066c8d339f041c9ac5ae68762a342d5d86/src/traces/box/cross_trace_calc.js#L103-L104) parameter.

I like this idea.

nicolaskruchten:
vote to close... will be covered in the v4 docs

PR #1098, opened 2018-09-21
Footersync

jdamiba:
partially addresses https://github.com/plotly/plotly-next/issues/159

matched footer with plot.ly

screen shot 2018-09-21 at 12 05 25 pm


nicolaskruchten:
💃

nicolaskruchten:
Ah actually... Could we standardize the casing? on the main site it's all all-caps

jdamiba:
Should be in all caps now!

Issue #1100, opened 2018-09-23
Update windrose examples to use barpolar


jonmmease:
Plotly.js 1.41 will be included in the plotly.py 3.3 release next week and it would be nice to have some `barpolar` documentation examples to reference in the release announcement.

To start with, would anyone have time to port the JavaScript and Python wind rose examples (https://plot.ly/python/wind-rose-charts/) from `area` to `barpolar`?

cc @etpinard @jackparmer

jackparmer:
@Kully or @michaelbabyn Can you please take this (whoever gets to it first)

The old `area` wind roses essentially shouldn't be in the docs anymore, they're deprecated with `barpolar`

Kully:
> @Kully or @michaelbabyn Can you please take this (whoever gets to it first)

I got it!

jonmmease:
It will be `Barpolar`, and hopefully I'll have a release candidate published by tomorrow morning.  And we'll need to run the examples before publishing right? So that the `barpolar` trace is displayed rather than `area`.  

Thanks @Kully!

jonmmease:
Here's the branch with barpolar in it https://github.com/plotly/plotly.py/pull/1202.  I'm hitting an issue using barpolar with FigureWidget that I need to work through before we release this but it's in fine shape for the docs.

jonmmease:
@Kully 3.3.0rc1 with barpolar support has been published to PyPI and anaconda cloud

PR #1101, opened 2018-09-27
area -> barpolar in python/js docs

Kully:
re: https://github.com/plotly/documentation/issues/1100


Kully:
@jonmmease

jonmmease:
Let's change the reference link at the bottom of the Python notebook to point to the `barpolar` attributes instead of `area`. Otherwise 💃

PR #1121, opened 2018-10-08
Fix cosmetic typo in example

gdw2:


cldougl:
thanks for the correction @gdw2

PR #1123, opened 2018-10-09
Add GitHub stars To Homepage

jdamiba:
screen shot 2018-10-09 at 4 20 58 pm

Part of https://github.com/plotly/plotly-next/issues/159

### DIFF
--- a/_data/cache_bust_css.yml
+++ b/_data/cache_bust_css.yml
@@ -1,2 +1,2 @@
-299bf4ed0ab496795b8fe3588fa04c34  ../all_static/css/main.css
+e1f40e63b86de7c2e0444738a2c8f9df  ../all_static/css/main.css
 a5765df8bc2358dd61c19edcc0cbef7d  ../all_static/css/normalize.css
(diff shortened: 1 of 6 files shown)


jdamiba:
I would like to add some space between the stars and the name of the library, but I'm having trouble getting css styles to load.

jdamiba:
screen shot 2018-10-10 at 9 47 50 am

I put the github stars into their own div and added a css class to `_home.scss` nested in the `category-wrap` class. This increased the space between between the name of the library and the github stars as desired but now the stars are in the bottom left of the div instead of the center. 

It seems that the styles in the github-stars CSS class are not being applied- the extra space is as a result of the markup.

jdamiba:
screen shot 2018-10-10 at 12 48 40 pm
screen shot 2018-10-10 at 12 48 44 pm

cldougl:
(fyi @nicolaskruchten the class issue got resolved if you're just following this thread via email )

nicolaskruchten:
Poor Matlab and its dismal star score :)

PR #1125, opened 2018-10-10
fixing warning

gdw2:
DeprecationWarning:
object of type <class 'float'> cannot be safely interpreted as an integer.


cldougl:
thanks for contributing @gdw2 !

PR #1133, opened 2018-10-16
fix syntax error in first plot and add font settings section to dashboard api

Kully:
in response to: https://github.com/plotly/plotly.py/issues/1225#event-1906442020

- updates the html file for dashboards api legacy doc ('16' -> integer)
- adds a section about updating the font with some examples

I did not:
- update the png image of the dashboard in the doc found here: https://plot.ly/python/create-online-dashboard/#upload-dashboard

The new dashboard with the modified font looks like: https://plot.ly/dashboard/PythonPlotBot:540/view#/

Don't think it is worth to update, especially since this is legacy and since we want to encourage users to check out Dash anyways. But I felt this update is useful for users like the one above who don't know that fonts are capable of being specified through the api.

### DIFF
--- a/_posts/python/legacy/dashboard/2015-06-30-dashboard-api.html
+++ b/_posts/python/legacy/dashboard/2015-06-30-dashboard-api.html
@@ -8,9 +8,9 @@ layout: user-guide
 name: Dashboard API
 language: python
 title: Dashboard API | plotly
-display_as: legacy_charts
+display_as: file_settings
 has_thumbnail: true
-page_type: u-guide
+page_type: example_index
 order: 0
 ipynb: ~notebook_demo/148
 ---
@@ -53,7 +53,7 @@ Run  <code>pip install plotly --upgrade</code> to update your Plotly version.</p
 
 
 <div class="output_text output_subarea output_execute_result">
-<pre>&#39;2.5.1&#39;</pre>
+<pre>&#39;3.3.0&#39;</pre>
 </div>
 
 </div>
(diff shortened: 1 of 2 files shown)


Kully:
@cldougl ready for a review

_posts/python/legacy/dashboard/2015-06-30-dashboard-api.html
```diff
@@ -8,9 +8,9 @@
 name: Dashboard API
 language: python
 title: Dashboard API | plotly
-display_as: legacy_charts
+display_as: file_settings
```
cldougl: @Kully was this an intended change - didn't we move this to legacy_charts on purpose?

_posts/python/legacy/dashboard/2015-06-30-dashboard-api.html
```diff
@@ -8,9 +8,9 @@
...
 has_thumbnail: true
-page_type: u-guide
+page_type: example_index
```
cldougl: same as above

_posts/python/legacy/dashboard/2015-06-30-dashboard-api.html
Kully: that was not intended - i wonder how that happened

perhaps running the publisher cell did it

Kully:
comments addressed:
  I ran `jekyll serve` to make sure - the doc is not on the plot.ly/python page and back in legacy

PR #1134, opened 2018-10-16
Add Mailchimp form to footer

jdamiba:
closes https://github.com/plotly/streambed/issues/11486#issuecomment-430242427

screen shot 2018-10-16 at 11 08 53 am
screen shot 2018-10-16 at 11 09 18 am
screen shot 2018-10-16 at 11 09 33 am


nicolaskruchten:
Could we have a border around the input box, and remove the `:` from the placeholder please? Would be good to remove the colon from the placeholder in the next repo also, I didn't see it there till just now.

nicolaskruchten:
Could we also have it be a sixth column in wide mode, rather than wrap below like in the first screenshot?

jdamiba:
Updated with @nicolaskruchten's styling suggestions. 

screen shot 2018-10-16 at 12 06 06 pm

jdamiba:
Updated to reflect @cldougl's review. Added padding to the text input box and added "target="_blank" to the form. 

screen shot 2018-10-16 at 12 36 02 pm

cldougl:
nice!!
_very_ nit-picky detail but I think it might look slightly nicer with a bit of top margin on the input box (I think it looks a little close to the text and spreading that out a tiny bit may look a little cleaner)
what do you think @jdamiba  ?

pad

jdamiba:
I agree, a bit of padding makes it easier on the eyes. 

screen shot 2018-10-16 at 12 43 09 pm

cldougl:
nice - that's all the feedback from me 
💃

PR #1145, opened 2018-10-22
Fixing broken links

jdamiba:
closes https://github.com/plotly/plotly-next/issues/181


michaelbabyn:
I had to fix the auto-commit (I wish there was a way to reference more than one line at a time) but now it's working as it should.

PR #1152, opened 2018-10-25
updating GTM tags

jdamiba:


_layouts/base.html
```diff
@@ -2,6 +2,8 @@
 <html>
 {% include head.html %}
 <body data-spy="scroll" data-target=".watch" style="position:relative;">
+	<noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-N6T2RXG"
```
nicolaskruchten: are there any other `<body>` tags?

_includes/head.html
```diff
@@ -35,16 +35,12 @@
...
-
     <!-- Global site tag (gtag.js) - AdWords: 1009791370 -->
-    <script async src="https://www.googletagmanager.com/gtag/js?id=AW-1009791370"></script>
-    <script>
-    window.dataLayer = window.dataLayer || [];
-    function gtag(){dataLayer.push(arguments);}
-    gtag('js', new Date());
-    gtag('config', 'AW-1009791370');
-    </script>
+    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
```
nicolaskruchten: are there any other `<head>` tags?

404.html
```diff
@@ -6,7 +6,20 @@
 <html>
 <head>
     {% include head.html %}
-
+<!-- Global site tag (gtag.js) - AdWords: 1009791370 -->
```
nicolaskruchten: The line above here actually includes the `_include` file below, so this would result in a double-addition of the tag.

nicolaskruchten: that said, now that I'm looking at this, the `head` include has the `<head>` tag so here we're nesting head in head which is bad. let's try to clean this up as part of this PR please.

_layouts/blog-post.html
```diff
@@ -2,10 +2,27 @@
 <html>
   {% include head.html %}
   <head>
-  <!-- <link rel="stylesheet" type="text/css" href="{{site.staticurl}}/css/state.css"> -->
+    <!-- Global site tag (gtag.js) - AdWords: 1009791370 -->
```
nicolaskruchten: Similar to the 404 comment... the `head` is being included here already so we don't want the tags again. BUT here we're including `head` and then redeclaring it, which is a no-no... let's fix this as well please.

_layouts/langindex.html
```diff
@@ -2,6 +2,10 @@
 <html>
 {% include head.html %}
 <body>
+    <!-- Google Tag Manager (noscript) -->
```
nicolaskruchten: So it looks lie we'll need this in many places. This is a good candidate for something that needs to be refactored into an include, so the tag itself is only in one place.

404.html
nicolaskruchten: It may be best to change `_includes.head.html` to NOT contain the `<head> ... </head>` tags, and go through and clean up the tags everywhere this include is loaded, so that 1) we never have containment or repetition issues and 2) the call-site has the option of adding extra stuff to the `<head>` tag.

404.html
```diff
@@ -6,11 +6,11 @@
 <html>
 <head>
     {% include head.html %}
-
     <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/js/bootstrap.min.js" integrity="sha384-0mSbJDEHialfmuBBQP6A4Qrprq5OVfW37PRR3j5ELqxss1yVqOtnepnHVP9aJ7xS" crossorigin="anonymous"></script>
```
nicolaskruchten: Ah, much clearer. So actually it seems like the only place we really try to add anything to the head tag is here... Is this necessary? Is this already in the head, or does the 404 look awful if we just didn't include this? We might be able to add the `<head>` tags back in and just remove this one line

_includes/google-tag-head.html
```diff
@@ -0,0 +1,14 @@
+<!-- Global site tag (gtag.js) - AdWords: 1009791370 -->
```
nicolaskruchten: good to pull this guy out, but it doesn't seem to be called from anywhere... can we not make use of it in some of the other places you've added this tag?

all_static/images/Plotly-Feed.html
```diff
@@ -4,6 +4,20 @@
 
 
 
+<!-- Global site tag (gtag.js) - AdWords: 1009791370 -->
```
nicolaskruchten: would be good to use an include here?

_includes/head.html
```diff
@@ -1,57 +1,60 @@
...
+
+<!-- LOCAL DEV STYLESHEET -->
+<!--<link rel="stylesheet" type="text/css" href="http://api.plotly.dev/all_static/css/main.css">-->
+
+<!-- Icon -->
+<link rel="shortcut icon" href="/images/plotly-ico.png" type="image/x-icon"/>
+
+<!-- Google Tags-->
+<!-- Global site tag (gtag.js) - AdWords: 1009791370 -->
+<script async src=""https://www.googletagmanager.com/gtag/js?id=AW-1009791370""></script>
```
nicolaskruchten: somehow you ended up with a double set of '"' here...

_includes/head.html
```diff
@@ -1,57 +1,60 @@
...
+<script>
+window.dataLayer = window.dataLayer || [];
+function gtag(){dataLayer.push(arguments);}
+gtag('js', new Date());
+gtag('config', 'AW-1009791370');
+</script>
+<!-- Google Tag Manager Tag -->
```
nicolaskruchten: this should be using the include from `google-tag-head.html`.

_includes/head.html
```diff
@@ -1,57 +1,60 @@
...
+<link rel="stylesheet" type="text/css" href="https://images.plot.ly/assets/css/normalize.css">
+<link rel="stylesheet" type="text/css" href="{{site.staticurl}}/css/main.css?version={{ site.data.cache_bust_css | truncate: 10, '' }}">
+
+<!-- LOCAL DEV STYLESHEET -->
+<!--<link rel="stylesheet" type="text/css" href="http://api.plotly.dev/all_static/css/main.css">-->
+
+<!-- Icon -->
+<link rel="shortcut icon" href="/images/plotly-ico.png" type="image/x-icon"/>
+
+<!-- Google Tags-->
```
nicolaskruchten: It seems like the whitespace in this file changed on every line, which makes reviewing challenging... could you please  undo this dedent?

404.html
```diff
@@ -4,13 +4,11 @@
 
 <!DOCTYPE html>
 <html>
-<head>
```
nicolaskruchten: nice cleanup

_layouts/blog-post.html
```diff
@@ -1,11 +1,10 @@
 <!DOCTYPE html>
 <html>
   {% include head.html %}
-  <head>
```
nicolaskruchten: good cleanup

_includes/head.html
```diff
@@ -1,48 +1,48 @@
...
-    <!--<link rel="stylesheet" type="text/css" href="http://api.plotly.dev/all_static/css/main.css">-->
+      <!-- Icon -->
+      <link rel="shortcut icon" href="/images/plotly-ico.png" type="image/x-icon"/>
 
-    <!-- Icon -->
-    <link rel="shortcut icon" href="/images/plotly-ico.png" type="image/x-icon"/>
+      <!-- Google Tags-->
+      {% include google-tag-head.html %}
+
+      {% include seo.html %}
```
nicolaskruchten: it looks like `seo` is getting included twice now... ?

nicolaskruchten:
phew, ok I think this guy's good 💃

PR #1171, opened 2018-11-09
use matrix for z input instead of vector

### LINKED ISSUE #1170: r contour example uses vector for z value instead of matrix
In this example https://plot.ly/r/contour-plots/#set-x-and-y-coordinates. Currently, Plotly is only using the first 5 elements of z (which should be a 5 x 5 matrix ).

michaelbabyn:
resolves #1170


cldougl:
💃

PR #1199, opened 2018-12-06
Version python docs

### LINKED ISSUE #990: Python: Version the docs with plotly.py
I would like to propose that we work towards hosting a build of the Python documentation for each released version of plotly.py, and add the version number to the URL.

This is the approach of the documentation for many other Python projects. See the pandas documentation for example: The documentation for a prior version of pandas can be accessed at https://pandas.pydata.org/pandas-docs/version/VERSION (e.g. https://pandas.pydata.org/pandas-docs/version/0.21.1/)

My thoughts is that we would continue to update the current URLs with the latest stable version.

Here are some motivations for the change:

 1. I would like to be able to update the legacy docs to use version 3 conventions (e.g. `go.layout.XAxis`), while still having a way for people to access the version 2.7 docs.

 2. I would like to work towards adding docstring hyperlinks to plotly.py that reference documentation pages, and with versioned docs these hyperlinks would always point to the page for the version of plotly.py that the person is actually using.

 3. I would like to get rid of the "Version Check" section at the top of all of the notebooks. It adds clutter and it wouldn't be needed if the docs were built for each version of plotly.py

How hard would this be from a documentation hosting point of view?

michaelbabyn:
This PR starts by updating some site logic so that we can have multiple versions of python without 2 of every post showing up on the index pages. It would resolve #990 

Then I went through to make sure all the syntax in the recent examples were plotly 2 compatible.

The links at the top of the sidebar point to the normal python docs for the moment.
side bar python

This includes the full reference guide so that will still include a recent version of the schema. It wouldn't be to complicated to generate a new guide specific to version 2 (version 1.38.0 of the plot schema) but I don't know enough about algolia search to set that up.

cc @cldougl


nicolaskruchten:
@michaelbabyn are you OK with closing this one out?

PR #1203, opened 2018-12-11
document virtualenv in getting started page

### LINKED ISSUE #947: Add jupyter and virtual environment install instructions on getting started page
we should add instructions and references for working with plotly in a jupyter notebook using virtual environments (for example installing jupyter and starting the kernal from the venv + a reference to troubleshoot common problems). 
This will be helpful for users that would like to use multiple versions (ie 2.x.x and 3.x.x) of plotly.py

michaelbabyn:
resolves #947

### DIFF
--- a/_posts/python/getting-started/getting-started.ipynb
+++ b/_posts/python/getting-started/getting-started.ipynb
@@ -45,7 +45,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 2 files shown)


_posts/python/getting-started/getting-started.ipynb
```diff
@@ -565,14 +657,32 @@
         }
        ],
        "layout": {
-        "title": "hello world"
+        "autosize": true,
```
cldougl: @michaelbabyn is this example code change meant to be in this pr ?
No problem if it's intentional

michaelbabyn: At some point while I was making this doc I re-ran the code block. I imagine the difference in the layout here has to do with the version of plotly I'm using.

PR #1213, opened 2018-12-18
fix broken plot for R/logos doc  

priyatharsan:


michaelbabyn:
looks good to me :dancer:

PR #1236, opened 2019-01-16
adding drift

jdamiba:
closes https://github.com/plotly/streambed/issues/12187


_config_dev.yml
```diff
@@ -1,7 +1,7 @@
 plugins:
-- jekyll-redirect-from
```
nicolaskruchten: please don't commit changes to this file...

nicolaskruchten:
should be fine once you revert the changes to the config file :) 💃

_config_dev.yml
jdamiba: Sorry, forgot that this repo has different linting rules than `next`. Reverted `config-dev.yml` to its original state.

PR #1238, opened 2019-01-17
fix ribbon plot example code

jacobq:
Set `trace1.type = 'surface'` on ribbon plot example. Without this there's an annoying, stray line plotted over the ribbon chart:
Also replaced tabs with spaces for consistency.

image


nicolaskruchten:
Thanks!

PR #1247, opened 2019-01-24
fix bug in js splom example by not recycling the same object for all axes

### LINKED ISSUE #1246: Splom documentation examples bug
The examples here https://plot.ly/javascript/splom/ suffer from https://github.com/plotly/plotly.js/issues/3466 because of the shared `axis` block. They should be edited to avoid leading people to do this :)

michaelbabyn:
resolves #1246. cc @nicolaskruchten


nicolaskruchten:
💃 if you can fix the CodePen as well please :)

michaelbabyn:
Cool, I did that too already.

PR #1252, opened 2019-02-01
Removed colorbar attribute from layout for scatter-map doc

priyatharsan:


_posts/python/maps/scatter-plot-on-maps/scatter-plot-on-map.ipynb
```diff
@@ -72,13 +99,12 @@
     "            color = df['cnt'],\n",
     "            cmax = df['cnt'].max(),\n",
     "            colorbar=dict(\n",
```
michaelbabyn: This isn't blocking the merge but we should try to replace the `dict()` method with the `go.Scattergeo()` whenever we see this in old docs now that validation works for this trace in `plotly.py`.

priyatharsan: Indeed! I forgot that. Its 👍 in https://github.com/plotly/documentation/pull/1252/commits/93f9002086a5ee70729bb293855d0928141e7730

Issue #1258, opened 2019-02-13
Latex Labels example is broken


Aypac:
There is an example of python syntax of labeling [here](https://plot.ly/python/LaTeX/). However, the latex is not rendered correctly:

![image](https://user-images.githubusercontent.com/10585956/52699295-b83a0500-2f75-11e9-9603-71b0bce14feb.png)

This relates to these file: https://github.com/plotly/documentation/tree/source-design-merge/_posts/python/style/latex

priyatharsan:
the rendering issue is due to a bug in plotly.js: https://github.com/plotly/plotly.js/issues/559, https://github.com/plotly/plotly.js/issues/3290 and not the documentation.

PR #1266, opened 2019-02-18
updated python latex doc

priyatharsan:
The plot may not render properly in FF due to this issue: https://github.com/plotly/plotly.js/issues/3290
and the hover label issue: https://github.com/plotly/plotly.js/issues/559

screenshot 8


michaelbabyn:
@priyatharsan, the code formatting looks better here than it did before the update but this PR doesn't really close #1258. Can you edit the first comment in this PR to remove the "closes 1258" reference and then manually close 1258 by citing that these are plotly.js bugs not documentation issues.

:dancer: for the PR though it's definitely an improvement

priyatharsan:
yeah that makes sense. Thanks for the review @michaelbabyn

PR #1269, opened 2019-02-26
Fix code indentation

clarkwinkelmann:
I just noticed this while browsing the docs.

I've updated the code but will clone and run jekyll shortly to check if it compiles.

EDIT: after waiting so long for the clone and forgetting about it, I finally tested my fix locally. There are many liquid syntax errors in the console but they are also there on the source branch and don't seem related to my fix.


michaelbabyn:
Thanks for the fix @clarkwinkelmann !

PR #1277, opened 2019-03-11
adding pardot tag

jdamiba:
addresses https://github.com/plotly/streambed/issues/12579


VeraZab:
should probably go to Google Tag Manager

PR #1283, opened 2019-03-22
Add bar plot

divyachandran-ds:


cldougl:
@divyachandran-ds we'll want to keep the gemfile.lock but it shouldn't need to be changed for this pr :)

PR #1312, opened 2019-04-20
Correction required in array slicing for frq & Y

ayulockin:
`frq = frq[range(n/2)]` and `Y = Y[range(n/2)]` showed `TypeError` with the following message:
`'float' object cannot be interpreted as an integer`. 
Since `150/2 = 75.0` which is a float and slicing require integer as index. Thus made relevant changes in `line 27 and 30`.


nicolaskruchten:
Thanks!

Issue #1317, opened 2019-04-25
update multiple chart type image


cldougl:
![Screen Shot 2019-04-25 at 4 07 54 PM](https://user-images.githubusercontent.com/2789078/56765126-67fcd180-6774-11e9-9f78-60f155e9ae92.png)

This should be an image of a chart with multiple trace types as in the example.

PR #1321, opened 2019-04-25
Thumbnail

Mahdis-z:


michaelbabyn:
Is this for #1317? If it is, you can add "resolves #1317" to the description above. That will make it clearer what issue this PR is addressing and it will also tell github to close the issue when this PR is merged.

PR #1323, opened 2019-04-26
reference side bar reorg

### LINKED ISSUE #1296: Reference page sidebar organization
The reference page sidebar should be reordered to match the trace-type selector in Chart Studio:

* Trace Type instead of Chart Type
* Sections for Simple, Distributions, 3D, Finance etc
* All trace types should be there including isosurface

nicolaskruchten:
closes #1296


nicolaskruchten:
This PR lists all trace types in the sidebar, organized like the trace type selector modal at https://plot.ly/create

In addition, it lists all the high-level objects in `layout` like `polar`, `ternary`, `images` etc at the expense of some overly-specific ones like `xaxis.title` etc.

PR #1324, opened 2019-04-26
Algolia version

cldougl:


nicolaskruchten:
@michaelbabyn this is resolved by some recent work you did right? If so, please close :)

PR #1329, opened 2019-05-02
Waterfall-Python

Mahdis-z:


Mahdis-z:
> Almost there but what I meant by section headers was just a title for each graph in a markdown cell like in [this example](https://github.com/plotly/documentation/blob/source-design-merge/_posts/python/basic/bubble/bubble.ipynb). We only need the version check and new to plotly cells once at the beginning of the document. After fixing that and re-running the publisher cell this should be good to go.

Ah, ok right!!

michaelbabyn:
The section headers are good now so the navigation side bar is properly filled out but I'm seeing the publisher cell in the doc
waterfall-docs-publisher-file
When we publish a ipynb file every cell except the last two are rendered into the html doc. In this case I think there is an extra publisher cell so that one of them is making it into the html file.

Mahdis-z:
> The section headers are good now so the navigation side bar is properly filled out but I'm seeing the publisher cell in the doc
> waterfall-docs-publisher-file
> When we publish an ipynb file every cell except the last two are rendered into the html doc. In this case I think there is an extra publisher cell so that one of them is making it into the html file.

Done done!

michaelbabyn:
Thanks for the changes @Mahdis-z ! This looks good to me. One thing I should note about these docs is to try and use consistent coding style throughout the examples. Generally we conform to [pep 8](https://www.python.org/dev/peps/pep-0008/#pet-peeves) with respect to how much space we put around `=` in various situations although I know this can be difficult working from within an ipynb.

Mahdis-z:
> Thanks for the changes @Mahdis-z ! This looks good to me. One thing I should note about these docs is to try and use consistent coding style throughout the examples. Generally we conform to [pep 8](https://www.python.org/dev/peps/pep-0008/#pet-peeves) with respect to how much space we put around `=` in various situations although I know this can be difficult working from within an ipynb.

Hmmm I see, okay there we go!!

nicolaskruchten:
Neat thing about the Jupytext doc flow: we can autoformat our notebook code at generation-time!

michaelbabyn:
This looks good to me @Mahdis-z :dancer:

PR #1339, opened 2019-05-18
remove b tags

nicolaskruchten:
These `<b>` tags are showing up in the actual page output and should be removed. I've already patched the production docs in ed52707af818cd9b9b8975d97025adda54c0a043


michaelbabyn:
:dancer:

PR #1356, opened 2019-06-12
update navbar. currently the style is updated via main.css directly

### LINKED ISSUE #1355: Navbar changes
These changes should only take 20min but I've spent an hour trying to install gems and ruby etc locally :(

I'd like to make the following simple changes to the navbar to somewhat align it to the main plotly site navbar.

image

michaelbabyn:
The sccs will need to be updated as well to make this permanent. Resolves #1355 

navbar-update


nicolaskruchten:
The differences between this and the asks in #1355 are fine with me given the expected vs realized level of effort :)

nicolaskruchten:
@cldougl I think you'll probably need to review this, as I worked on it a lot with Micheal so my review would be biased :P

PR #1370, opened 2019-06-21
change redraw to true for choropleth map animation

### LINKED ISSUE #1365: Map animation not uploading
https://plot.ly/javascript/map-animations/

The map animation isn't displaying the updated map or tooltips when "play" or a different year is selected, but it correctly updates when the map is double clicked. This bug seems to be appearing everywhere, including the codepen example.

michaelbabyn:
resolves #1365

### DIFF
--- a/_posts/plotly_js/animations/map-animations/2017-08-22-choropleth-animation.html
+++ b/_posts/plotly_js/animations/map-animations/2017-08-22-choropleth-animation.html
@@ -1,6 +1,6 @@
 ---
 name: Map Animations
-plot_url: https://codepen.io/bcd/embed/rzrJjb/?height=500&theme-id=15263&default-tab=result
+plot_url: https://codepen.io/plotly/embed/YoZLGQ/?height=500&theme-id=15263&default-tab=result
 language: plotly_js
 suite: map-animations
 order: 1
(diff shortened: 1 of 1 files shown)


_posts/plotly_js/animations/map-animations/2017-08-22-choropleth-animation.html
```diff
@@ -1,6 +1,6 @@
 ---
 name: Map Animations
-plot_url: https://codepen.io/bcd/embed/rzrJjb/?height=500&theme-id=15263&default-tab=result
+plot_url: https://codepen.io/plotly/embed/YoZLGQ/?height=500&theme-id=15263&default-tab=result
```
cldougl: @michaelbabyn is this the correct codepen link? https://codepen.io/plotly/embed/YoZLGQ/?height=500&theme-id=15263&default-tab=result it doesn't seem to be animating to me?

michaelbabyn: Yeah, it is. The data doesn't change very much year to year but you should notice that it works if you go from 2016 to 2010.

cldougl: do you think we could use different data that does visibly change to make this example a bit stronger? (even hovering on the data it didn't really look like the #s were changing).

michaelbabyn: Yes, I think that would be a much better fix.

michaelbabyn:
I switched the dataset to gapminder here to increase the contrast in the animation and there was some other stuff wrong with this example that was preventing the hover from changing that is also fixed now.

cc @cldougl

PR #1385, opened 2019-07-04
Fix bad tags

nicolaskruchten:


_posts/ggplot2/2016-12-16-ggplot2-index.md
```diff
@@ -14,7 +14,7 @@ redirect_from: ggplot2/reference/
       <div class="--category-img"><img src="https://images.plot.ly/language-icons/api-home/ggplot2-logo.png" alt=""></div>
       <div class="--body">
         <h1>Plotly ggplot2 Library</h1>
-        <p>{{page.description}}</consectetur>
+        <p>{{page.description}}</p>
         </p>
```
michaelbabyn: It looks like this closing p tag below the consectetur tag is functional so we should just delete the `</consectetur>` instead of replacing it with a p.

nicolaskruchten:
@michaelbabyn good catch! I flattened out the duplicated p tags :)

PR #1388, opened 2019-07-04
resolve liquid warnings

nicolaskruchten:


nicolaskruchten:
I'm surprised at how lax Liquid is that these were just warnings ;)

PR #1394, opened 2019-07-09
v4 pre-release warning

nicolaskruchten:


nicolaskruchten:
Looks like this: 
image

PR #1398, opened 2019-07-10
python user guide modification

Mahdis-z:
closes https://github.com/plotly/documentation/issues/1225 and https://github.com/plotly/documentation/pull/1273


_posts/python/fundamentals/user_guide_python/new_py_guide.ipynb
```diff
@@ -1594,41 +14443,56 @@
...
+     "execution_count": 88,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
    "source": [
-    "del fig.layout.annotations[0]   #deletes annotation for `First Subplot`\n",
-    "del fig.layout.annotations[0]   #deletes annotation for `Second Subplot` because of shift\n",
-    "fig.layout.xaxis1.domain = [0.0, 1]"
+    "fig.layout.annotations = []  #deletes annotation for `First Subplot`\n",
```
michaelbabyn: This will actually delete all three annotations. The intent is to only delete the first two so it should be something like this which will replace the three annotations with only the last one.
```suggestion
    "fig.layout.annotations = [fig.layout.annotations[2]]  #deletes annotations for `First Subplot` and `Second Subplot`\n",
```

nicolaskruchten:
heads-up that we shouldn't really touch the Python docs until we do the v4 cutover, as none of these changes will be ported... is this PR responding to a particular issue?

PR #1419, opened 2019-07-27
typecast map to list in cmocean notebook

srcoulombe:
Fixes typo/missing typecast that otherwise raises TypeError: 'map' object is not subscriptable


srcoulombe:
plotly already supports cmocean colormaps, this pull request is obsolete

PR #1434, opened 2019-08-01
added geom_hex

bobidou23:


_posts/ggplot2/2019-07-30-geom_hex.Rmd
```diff
@@ -0,0 +1,147 @@
...
+chart_link
+```
+
+### Customized Colours
+Let's flip the colour scheme so that lighter colours denote larger numbers than darker colours. We should also move to a logarithmic scale, since as it is, the very large value in the bottom right overshadows all other values.
+
+```{r, results='hide'}
+library(plotly)
+
+p <- ggplot(english_french, aes(x=engperc,y=frenperc)) +
```
michaelbabyn: It would be nice if `english_french` was assigned in this example too so that the code here could run independently of the first one.

_posts/ggplot2/2019-07-30-geom_hex.Rmd
```diff
@@ -0,0 +1,147 @@
...
+chart_link
+```
+
+### Weighted Data
+In the previous graphs, each observation represented a single census subdivision - this counted small towns of 500 people equally with cities like Montreal and Toronto. We can weight the data by the "total" column (i.e. total population) to make this a graph of population.
+
+```{r, results='hide'}
+library(plotly)
+
+p <- ggplot(english_french, aes(x=engperc, y=frenperc, weight=total)) +
```
michaelbabyn: same comment as above re: `english_french`

_posts/ggplot2/2019-07-30-geom_hex.Rmd
```diff
@@ -0,0 +1,147 @@
...
+chart_link
+```
+
+### Customized Appearance
+We can modify the graph's appearance - for example, if the grey background makes it difficult to make out the paler shades of blue, we can change the theme to one with a white background. Included also is a way to change the font. You can find a list [here](http://ggplot2.tidyverse.org/reference/theme.html) of all the theme elements that you can modify.
+
+```{r, results='hide'}
+library(plotly)
+
+p <- ggplot(english_french, aes(x=engperc,y=frenperc, weight=total)) +
```
michaelbabyn: again, if we could assign `english_french` in this example before using it, that would be great.

michaelbabyn:
Thanks @bobidou23 , feel free to merge this! :dancer:

PR #1435, opened 2019-08-01
added geom_text

bobidou23:


_posts/ggplot2/2019-07-30-geom_text.Rmd
```diff
@@ -0,0 +1,193 @@
...
+chart_link = api_create(p, filename="geom_text/overlaid-points")
+chart_link
+```
+
+```{r echo=FALSE}
+chart_link
+```
+
+### Customed Colour and Size Scale
+Let's use the LaCroixColoR package to spruce up the colour scheme. In addition, by using scale_size_continuous, we can make sure that none of the text is too small.
```
michaelbabyn: The underscores in scale_size_continuous are causing some formatting issues instead of being rendered:

Screenshot from 2019-08-12 16-00-34

I think that putting it inside backticks should fix it: `scale_size_continuous`

_posts/ggplot2/2019-07-30-geom_text.Rmd
```diff
@@ -0,0 +1,193 @@
...
+chart_link = api_create(p, filename="geom_text/customized-scales")
+chart_link
+```
+
+```{r echo=FALSE}
+chart_link
+```
+
+### Adding a regression
+Adding a regression line as well as a label. geom_smooth does not allow for adjusting the transparency of the line (using alpha), which is why stat_smooth is used here. annotate is used to include a single text label (geom_text would create one label for every data point, all overlapped with each other).
```
michaelbabyn: can we put all function names inside backticks? (`stat_smooth`, `geom_text`, `annotate` ...)

Screenshot from 2019-08-12 16-05-58

_posts/ggplot2/2019-07-30-geom_text.Rmd
```diff
@@ -0,0 +1,193 @@
...
+```
+
+### Overlaid Points
+Colour-coding the text itself might present readability issues. Another possible use of geom_text is to keep the text grey, but overlay it on a coloured point graph.
+
+Adding the *text* option within aes() allows us to control the text that appears when hovering over a point.
+
+```{r, results='hide'}
+library(plotly)
+
```
michaelbabyn: Can we add dataframe assignment here so that the example runs without having run the first example first.

```r
recent_turnout <- read.csv("https://raw.githubusercontent.com/plotly/datasets/master/european_turnout.csv",stringsAsFactors = FALSE)
recent_turnout$region <- factor(recent_turnout$region, levels=c("British","Northern","Western","Mediterranean","Central/Eastern"))
```

PR #1442, opened 2019-08-06
Waterfall funnel

Mahdis-z:


_posts/plotly_js/financial/funnel/2019-06-27-funnel_plotly_js_index.html
```diff
@@ -3,12 +3,13 @@
 name: Funnel and Funnelarea Charts
 permalink: javascript/funnel-charts/
 description: How to make a D3.js-based funnel chart in javascript.
-layout: user-guide
+layout: base
 thumbnail: thumbnail/funnel.jpg
 language: plotly_js
+page_type: example_index
```
michaelbabyn: We've got a lot of financial charts on the homepage now so I think we should create an [index page](https://github.com/plotly/documentation/blob/source-design-merge/_posts/plotly_js/basic/2017-02-24-plotly_js-basic-index.html) like we have for some of the other sections so we only show five thumbnails from this section. 

Personally, I'd like to keep candlestick, indicators, waterfalls, and funnels here and move the others to the index page.

_posts/plotly_js/financial/indicator1/2015-12-20-indicator_plotly_js_index.html
```diff
@@ -3,7 +3,7 @@
 name: Indicators
 permalink: javascript/indicator/
 description: How to make a D3.js-based gauge chart in javascript.
-layout: user-guide
+layout: base
```
michaelbabyn: The layout attribute should be set to user-guide for all the docs in the financial section now so that they all appear in the index page. This is how it works for all the other index pages.

PR #1443, opened 2019-08-06
made a new version of geom_jitter

bobidou23:
not an integral part of my project

realized that the dataset I used to make geom_violin would be well-suited as well for geom_jitter, whose current page isn't super-descriptive. I made sure to include sections overlaying boxplots and text (like in the original version) but couldn't figure out how to incorporate error bars well.


_posts/ggplot2/2019-08-06-geom_jitter.md
```diff
@@ -0,0 +1,208 @@
...
+        plot.title = element_text(hjust = 0.5)) +
+  labs(title = "Democratic performance in the 2018 House elections, by region and density",
+       x = "Density Index from CityLab",
+       y = "Democratic Margin of Victory/Defeat") +
+  theme(text = element_text(family = 'Fira Sans'))
+ggplotly(p, tooltip=c("district","y"))
+
+# Create a shareable link to your chart
+# Set up API credentials: https://plot.ly/r/getting-started
+chart_link = api_create(p, filename="geom_jitter/jitterdodge")
```
michaelbabyn: The custom tooltip isn't making it to the chart we upload to the api.
Screenshot from 2019-08-30 09-15-00
You should either update p or use `ggplotly(p, tooltip=c("district","y"))` as an argument in api_create. I think updating p is more consistent with the rest of the docs though (`p <- ggplotly(p, tooltip=c("district","y"))`).

bobidou23: I fixed this!

PR #1445, opened 2019-08-07
new geom!

bobidou23:
should be a final version!


_posts/ggplot2/2019-08-06-geom_density_2d.Rmd
```diff
@@ -0,0 +1,166 @@
+---
+title: geom_density_2d | Examples | Plotly
+name: geom_density_2d
```
michaelbabyn: Even though all the other names in this section have the underscore format I think we should break this name up because it's a bit too long for its container. Maybe `geom density_2d`?

Screenshot from 2019-08-16 17-10-58

PR #1449, opened 2019-08-09
geom_count

bobidou23:
using flight data and beer data


_posts/ggplot2/2019-08-08-geom_count.Rmd
```diff
@@ -0,0 +1,177 @@
...
+
+```{r}
+library(plotly)
+packageVersion('plotly')
+```
+
+### Basic geom\_count Plot
+geom\_count is a way to plot two variables that are not continuous. Here's a modified version of the nycflights13 dataset that comes with R; it shows 2013 domestic flights leaving New York's three airports. This graph maps two categorical variables: which of America's major airports it was headed to, and which major carrier was operating it.
+
+It's good to show the ful airport names for destinations, rather than just the airport codes. You can use aes(group = ), which doesn't modify the graph in any way but adds information to the labels.
```
michaelbabyn:
```suggestion
It's good to show the full airport names for destinations, rather than just the airport codes. You can use aes(group = ), which doesn't modify the graph in any way but adds information to the labels.
```

PR #1454, opened 2019-08-12
geom_raster

bobidou23:
again only two images, one with the default colour scheme and one with a customized one. Maybe I could annotate text or something? (It would be very basic, given how little I know about the subject matter)

### DIFF
--- a/_posts/ggplot2/2019-08-12-geom_raster.Rmd
+++ b/_posts/ggplot2/2019-08-12-geom_raster.Rmd
@@ -43,9 +43,10 @@ geom\_raster creates a coloured heatmap, with two variables acting as the x- and
 
 ```{r, results='hide'}
 library(reshape2)
+library(plotly)
+
 df <- melt(volcano)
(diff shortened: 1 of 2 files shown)


_posts/ggplot2/2019-08-12-geom_raster.Rmd
```diff
@@ -0,0 +1,94 @@
...
+```
+
+### Basic 2d Heatmap
+geom\_raster creates a coloured heatmap, with two variables acting as the x- and y-coordinates and a third variable mapping onto a colour. (It is coded similarly to geom\_tile and is generated more quickly.) This uses the volcano dataset that comes pre-loaded with R.
+
+```{r, results='hide'}
+library(reshape2)
+df <- melt(volcano)
+
+library(plotly)
```
michaelbabyn: Usually we keep all the package imports at the top of the example. I think it's best to start with `library(plotly)` as well.

PR #1464, opened 2019-08-19
js density mapbox

Mahdis-z:


_posts/plotly_js/maps/density-mapbox/2019-08-16-multi_density-mapbox.html
```diff
@@ -0,0 +1,22 @@
...
+language: plotly_js
+suite: density-maps
+order: 2
+sitemap: false
+arrangement: horizontal
+
+---
+var data = [
+  {type: "densitymapbox", lon: [10, 20, 30], lat: [15, 25, 35], z: [1, 3, 2],
+   zmid: 0, radius: 50, colorbar: {y: 1, yanchor: 'top', len: 0.45}},
```
michaelbabyn: I find the `zmid: 0` here confusing since I don't know what negative density represents and, based on this output, it looks like plotly doesn't know either. Most of the time people want to use zmin/zmax to make sure that two colorscales agree despite having different data so I think that would be a better use case here (e.g we can set zmin,zmax to 0,5 for both traces).

_posts/plotly_js/maps/density-mapbox/2019-08-16-earthquack_density.html
```diff
@@ -0,0 +1,28 @@
...
+arrangement: horizontal
+
+---
+Plotly.d3.csv('https://raw.githubusercontent.com/plotly/datasets/master/earthquakes-23k.csv',
+  function(err, rows){function unpack(rows, key) {return rows.map(function(row){ return row[key];
+})};
+
+var data = [{
+  lon: unpack(rows, 'Longitude'), lat: unpack(rows, 'Latitude'), radius:10,
+  z: unpack(rows, 'quakes.Magnitude'), type: "densitymapbox", coloraxis: 'coloraxis',
```
michaelbabyn: Right now the colorbar is messed up because all the z data is undefined.
```suggestion
  z: unpack(rows, 'Magnitude'), type: "densitymapbox", coloraxis: 'coloraxis',
```

_posts/plotly_js/maps/density-mapbox/2019-08-16-earthquack_density.html
```diff
@@ -0,0 +1,28 @@
...
+  hoverinfo: 'skip'}];
+
+var layout = {
+    mapbox: {center: {lon: 60, lat: 30}, style: "outdoors", zoom: 2},
+    coloraxis: {colorscale: "Viridis"}, title: {text: "Earthquake Magnitude"},
+    width: 600, height: 400, margin: {t: 30, b: 0}};
+
+var config = {mapboxAccessToken: 'pk.eyJ1IjoiZXRwaW5hcmQiLCJhIjoiY2luMHIzdHE0MGFxNXVubTRxczZ2YmUxaCJ9.hwWZful0U2CQxit4ItNsiQ'};
```
michaelbabyn: We should replace the actual token here with `mapboxAccessToken: 'enter access token here'` to make it clear that we want people to get their own.

_posts/plotly_js/maps/2019-08-12-plotly_js-map-index.html
```diff
@@ -0,0 +1,30 @@
+---
+permalink: javascript/mapbox-charts/
```
michaelbabyn: This should be "maps" in order to agree with python and r
```suggestion
permalink: javascript/maps/
```

_posts/plotly_js/maps/2019-08-12-plotly_js-map-index.html
```diff
@@ -0,0 +1,30 @@
...
+title: Plotly.js Mapbox Charts
+description: Plotly.js makes interactive, publication-quality graphs online. Examples of how to make mapbox charts.
+name: More Mapbox Charts
+layout: langindex
+language: plotly_js
+display_as: maps
+has_thumbnail: true
+thumbnail: thumbnail/mixed.jpg
+page_type: example_index
+order: 5
```
michaelbabyn: I think we need a higher number than 5 here because this thumbnail isn't appearing at the end.

Screenshot from 2019-08-21 14-50-09

michaelbabyn: Could we swap these two scatter plot thumbnails with the bubble and the lines? I think that would better demonstrate the different types of traces we can make.

PR #1485, opened 2019-08-29
match permalink for js/py mapbox

Mahdis-z:


michaelbabyn:
:sweat_smile: , sorry I missed that! This looks good :dancer:

PR #1486, opened 2019-08-30
updated geom_quantile!

bobidou23:
better examples + explanation

### DIFF
--- a/_posts/ggplot2/2017-04-21-geom_quantile.md
+++ b/_posts/ggplot2/2017-04-21-geom_quantile.md
@@ -36,15 +36,55 @@ packageVersion('plotly')
 ```
 
 ```
-## [1] '4.9.0'
+## [1] '4.8.0.9000'
 ```
 
 ### Basic Example
+While common linear regression is a method of estimating the conditional mean of variable y based on the values of variable(s) x, quantile regression is a method that can give the conditional median (50th percentile) as well as any other quantile.
+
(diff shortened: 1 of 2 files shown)


_posts/ggplot2/2017-04-21-geom_quantile.md
```diff
@@ -70,6 +73,44 @@ chart_link
...
+  ptd <- factor(ptl > 0)
+  ftv <- factor(ftv)
+  levels(ftv)[-(1:2)] <- "2+"
+  data.frame(low = factor(low), age, lwt, race, smoke = (smoke > 0),
+             ptd, ht = (ht > 0), ui = (ui > 0), ftv, bwt)
+})
+
+p <- ggplot(df, aes(lwt, bwt, colour=smoke)) +
+  geom_point(size = 1) +
+  geom_quantile(quantiles = c(0.1, 0.5, 0.9), size = 2, alpha = 0.5)
```
michaelbabyn: I really like the idea of using multiple quantiles for each smoker category but I find this graph a little hard to read since all the quantiles regression lines look the same. Do you think we can we use the `..quantile..` computed variable in an `aes` here to visually separate them?

bobidou23:
wanted to differentiate the different quantiles by line type but that didn't seem to be possible because the ..quantile.. measure was continuous?

Anyway, restricted only to non-smokers and coloured by quantile:

スクリーンショット 2019-08-30 午後3 05 19

michaelbabyn:
You can make a continuous variable discrete using the `as.factor()` function (eg `aes(linetype = as.factor(..quantile..))`)  but I think setting the alpha to be a quantile would look good in this case (`aes(alpha = ..quantile..)`) although you'd probably need to scale the alpha if you went that route.

Issue #1412, opened 2019-07-25
1.49 Mapbox features


nicolaskruchten:
@etpinard fill in some links to mocks, codepens please

etpinard:
### `choroplethmapbox`

- Simple mock: https://rreusser.github.io/plotly-mock-viewer/#mapbox_choropleth0
- Mock with `geojson` set as `{/* geojson object /*}`: https://rreusser.github.io/plotly-mock-viewer/#mapbox_choropleth-raw-geojson
- Mock with multiple `choroplethmapbox` traces: https://rreusser.github.io/plotly-mock-viewer/#mapbox_choropleth-multiple

### `densitymapbox`

- Simple mock: https://rreusser.github.io/plotly-mock-viewer/#mapbox_density0
- Multiple `densitymapbox` traces: https://rreusser.github.io/plotly-mock-viewer/#mapbox_density-mulitple
- `densitymapbox` with "real" data: https://rreusser.github.io/plotly-mock-viewer/#mapbox_earthquake-density (essentially a remake of https://docs.mapbox.com/mapbox-gl-js/example/heatmap-layer/)

### New `mapbox.style` values

- `open-street-map`: https://rreusser.github.io/plotly-mock-viewer/#mapbox_osm-style
- `staman-*` styles: https://rreusser.github.io/plotly-mock-viewer/#mapbox_stamen-style
- `carto-*` styles: https://rreusser.github.io/plotly-mock-viewer/#mapbox_carto-style

nicolaskruchten:
New JS pages, say 2 examples each (simple and complicated): lines on a mapbox, filled areas on a mapbox, density on a mapbox, choropleth on a mapbox

New JS page: mapbox overview...
   * explanation of layers (style, data, layers), explanation of relationship to Mapbox.js vs Mapbox tile service
   * clear explanation of when a mapbox token is and is not required, and how to set it (either in the mapbox layout object or via configuration)
   * explanation of attribution
   * example of various styles (non-mapbox, mapbox)
   * links to the 5 data pages (scatter, line, area, choropleth, density)
   * examples of layers (environment Canada layer) based on https://codepen.io/nicolaskruchten/pen/BgbxQG?editors=0010

Mahdis-z:
- density mapbox https://github.com/plotly/plotly.js/pull/3993
- coropleth mapbox: https://github.com/plotly/plotly.js/pull/3988
[...]

PR #1490, opened 2019-09-03
define line and area on mapbox [JS]

Mahdis-z:
closes https://github.com/plotly/documentation/issues/1412


nicolaskruchten:
* section/example headers need to be descriptive of the examples: less emphasis on tiles and tokens outside of the "layers" page
* more explanatory text around "there are N ways to do this, here is how they compare, here are links to the other ways, here is one example of each"
* Line and area should be separate pages
* While we're at it we likely need "filled areas on maps" (non-mapbox)
* Line: we should document the "broken line" trick
* Area: we should document the "broken trace" trick and we should point out that there are basically three ways in Mapbox to make areas... fill-to-self, geojson layers, and the new choroplethmapbox trace type
* for the non-mapbox case there is only the one way to make filled areas today.

note: let's start with the Python versions of these pages please

nicolaskruchten:
Please rebase onto s-d-m- to get CI up and running :)

nicolaskruchten:
Should match the permalinks and contents of https://github.com/plotly/plotly.py-docs/pull/125

PR #1510, opened 2019-09-23
removing unused includes 

jdamiba:
This PR removes unused/unreferenced files from the `_includes` folder in this repo.

In order to accomplish this, it was necessary to modify the remaining `dashboard` pages that were not addressed by https://github.com/plotly/documentation/commit/098e17661c553ba1c2294ba6b9f46c79958517bd. 

I am reasonably confident that this is safe given that the Jekyll application builds locally when I don't `exclude` any files in the `config.yml`.  If there was a `layout` or `post` which still tried to reference these deleted `includes`, then Jekyll would throw a build-time error, which is in fact how I discovered the remaining dashboard pages.


nicolaskruchten:
Can you rebase this onto `source-design-merge` (aka SDM) please?

PR #1512, opened 2019-09-23
Revmoving unused includes and layout

jdamiba:


nicolaskruchten:
This is a shorter list than expected... both footer and footer-new (or whatever) are in use? I thought we would be able to drop some additional header/footer includes

nicolaskruchten:
Interestingly enough, `head.html` is included from `404.html`, so maybe Jekyll is too nice about not complaining about missing include files??

nicolaskruchten:
Ah ignore that last comment... I got mixed up between head and header.html!

404.html
```diff
@@ -15,7 +15,7 @@
     your browser</a> to improve your experience.</p>
 <![endif]-->
 
-{% include lang-splash-banner.html %}
+<div style="margin-top: 80px"></div>
```
nicolaskruchten: somehow my previous comment didn't make it to the right place... instead of inlining this `<div>` I would move it to the top of `_new/_page-components/_header-main.html`

PR #1513, opened 2019-09-24
Remove Unused Includes and Layout

jdamiba:
This PR deletes unused files in the `_includes` and `_layouts` folder in this repo. 

I believe this is safe because the Jekyll site still builds despite these files being deleted. If one of these files was referenced in another document, Jekyll would throw an error during build time.


nicolaskruchten:
💃

PR #1518, opened 2019-09-24
Refactor Includes Folder

jdamiba:
The purpose of this PR is to re-organize the `includes` folder in this repo. 

Going forward, the `includes` folder will have 3 subfolders:

`layouts` for includes that are used in layouts
`posts` for includes that are used in posts
`both` for includes that used in both layouts and posts

I believe that this re-organization of the `includes` folder is safe as Jekyll would throw a build-time error if any of the `includes` URLs was not correct.


nicolaskruchten:
:dancer:

PR #1519, opened 2019-09-25
Refactor Layouts Folder

jdamiba:
The purpose of this PR is to refactor the `layouts` folder in this repository so that all of the posts use either the `langindex` or the `base` layout. 

For the posts which are being transitioned from the `getstart-base` or `getstart` layout to the `base` layout it is necessary to change their `name` attribute to compensate for the fact that the `base` layout adds `in <lang>` to the end of the string.


nicolaskruchten:
to make it possible to review and merge this without too many sync problems with the `plotly.py-docs` repo, I would recommend restoring `_layouts/user-guide.html` as a copy of `_layouts/base.html` (and adjusting any `if`s that depend on `base` vs `user-guide` to generalize!) until this is all deployed, and then we can update the `plotly.py-docs` repo to use `base` instead of `user-guide`.

PR #1536, opened 2019-10-01
removing reference to unused search-suggestions.html include

jdamiba:
This PR removes an unused include.


nicolaskruchten:
Did you check Percy here?

jdamiba:
I have updated this PR to re-introduce the `ul` which the front-end uses to populate the side-bar. The Percy diffs have been investigated to make sure that the side-bar renders correctly. The remaining diffs are false positives.

nicolaskruchten:
💃

PR #1541, opened 2019-10-02
changing plotly.js to JavaScript in navigation

jdamiba:
The purpose of this PR is to change `plotly.js` to `JavaScript` in the navigation.


nicolaskruchten:
💃

Issue #1527, opened 2019-09-29
Reduce number of files loaded per page


nicolaskruchten:
if you view-source on a typical page you see that we're loading tons of extra files via `script` and `link` tags. Some are being loaded twice, some are being loaded for no reason. Google Fonts/font-awesome are being loaded via `link` *and* via `@import` in the CSS file itself.

In principle we should be able to reduce all the Javascript to a single `script src=` tag that loads the concatenation of all of the scripts we depend on (although loading common libs like jquery and bootstrap from external CDNs is probably all right) and we should definitely not load anything twice or anything unused.

PR #1547, opened 2019-10-10
Bring Prod and Dev CSS builds into alignment + reduce calls to external assets

jdamiba:
closes https://github.com/plotly/documentation/issues/1527

The purpose of this PR is to begin refactoring how the repo's CSS is generated. 

Rather than using the built CSS in `_site/all_static` in `dev` mode while using CSS fetched from `{{site.staticurl}}/css/main.css` in `production` and in CircleCi, this PR switches to using the `/all_static` URL, which is now proxy-passing the correct content due to changes in the nginx config for `streambed` (see https://github.com/plotly/streambed/pull/13611). 

At the same time, some duplication is being removed. Rather than making two calls to Google Fonts, we now make one, and some extraneous `@imports` have been deleted. The codehighlighting CSS and `normalize.scss` have been folded into the `main.scss` file, reducing the need for three `link` tags in the head. 

Note*: As a result of this PR, there are some slight changes to the way example code is highlighted. I believe that this is OK because the important thing about code-highlighting is that it happens. As long as semantically different pieces of example code still stand out as such, we should not block merging this PR because some of the highlight colors changed. It is highly unlikely that users depend on the specific color pallete used to highlight example code and notably, the code-highlight colors don't change for Python and R posts.


_config.yml
```diff
@@ -6,9 +6,9 @@ plugins:
 # ---
 # Define URLs
 # ---
-url: https://plot.ly
-BASE_URL: https://plot.ly
-staticurl: https://plot.ly/gh-pages/documentation/static/
+url: /
```
nicolaskruchten: so if these are all `/` ... can we get rid of these three config variables? are they required/used anywhere?

_includes/layouts/_footer-main.html
```diff
@@ -135,14 +135,14 @@ <h6 style="color:#119dff" class="--footer-heading">JOIN OUR MAILING LIST</h6>&#x
 <script src="//cdn.jsdelivr.net/algoliasearch.helper/2.1.0/algoliasearch.helper.min.js"></script>
 <script src="//cdn.jsdelivr.net/hogan.js/3.0.2/hogan.min.js"></script>
 <script src="//cdn.jsdelivr.net/momentjs/2.10.3/moment.min.js"></script>
-<script src="{{site.staticurl}}/javascripts/algolia/algolia.js"></script>
+<script src="/all_static/javascripts/algolia/algolia.js"></script>
```
nicolaskruchten: we can do this in another PR, but I'd love to merge these `all_static` JS blocks into on script :)

_includes/layouts/head.html
```diff
@@ -16,25 +16,18 @@
...
-      <link rel="stylesheet" type="text/css" href="https://images.plot.ly/assets/css/normalize.css">
-      <link rel="stylesheet" type="text/css" href="{{site.staticurl}}/css/main.css?version={{ site.data.cache_bust_css | truncate: 10, '' }}">
+      <link rel="stylesheet" type="text/css" href="/all_static/css/main.css?version={{ site.data.cache_bust_css | truncate: 10, '' }}">
 
       <!-- LOCAL DEV STYLESHEET -->
       <!--<link rel="stylesheet" type="text/css" href="http://api.plotly.dev/all_static/css/main.css">-->
 
       <!-- Icon -->
-      <link rel="shortcut icon" href="/images/plotly-ico.png" type="image/x-icon"/>
+      <link rel="shortcut icon" href="/all_static//images/plotly-ico.png" type="image/x-icon"/>
```
nicolaskruchten: double `//` here should be single

scss/main.scss
```diff
@@ -37,3 +35,594 @@ pre, code, .hljs{
 @import "layout/layout.grid";
 
 @import "layout/layout.upgrade-cta";
+
```
nicolaskruchten: so here instead of inlinig `normalize.css` why don't we just `include` it via SCSS so we remember that it's a standalone CSS file?

_includes/layouts/_footer-main.html
```diff
@@ -135,14 +135,14 @@ <h6 style="color:#119dff" class="--footer-heading">JOIN OUR MAILING LIST</h6>&#x
...
  <!-- Hash offset Script -->
-<script src="{{site.staticurl}}/javascripts/hash_offset.js"></script>
+<script src="/all_static//javascripts/hash_offset.js"></script>
  <!-- Image Hover Script -->
-<script src="{{site.staticurl}}/javascripts/imghover.js"></script>
+<script src="/all_static/javascripts/imghover.js"></script>
  {% if page.language != "python" %}
 <!-- code highlighting -->
-<script src="{{site.staticurl}}/javascripts/codehighlight/highlight.pack.js"></script>
+<script src="/all_static/javascripts/codehighlight/highlight.pack.js"></script>
```
nicolaskruchten: so there's another Highlight.js-related JS script load in the header that's also in an "if language is not python" block... can these be combined or something? It seems weird to have it in two places...

scss/main.scss
jdamiba: Sure, that makes sense. It would be easy to lose track of it otherwise.

_config.yml
jdamiba: Yes, these config variables are unnecessary given the `streambed` redirect for `/all_static`.

_includes/layouts/_footer-main.html
jdamiba: Sure- I've created an issue -> https://github.com/plotly/documentation/issues/1550

_includes/layouts/_footer-main.html
jdamiba: Will be addressed in https://github.com/plotly/documentation/issues/1550.

jdamiba:
@nicolas This PR has been updated per your latest comments.

scss/main.scss
```diff
@@ -37,3 +38,171 @@ pre, code, .hljs{
...
+
+
+/*
+
+Colorbrewer theme
```
nicolaskruchten: can we load this also via SCSS instead of inlining it? this is `tomorrow.css` right?

nicolaskruchten: sorry, this is `js-splash-highlight.css` right? What happened to `tomorrow.css`?

_config_python_search.yml
```diff
@@ -6,9 +6,9 @@ plugins:
 # ---
 # Define URLs
 # ---
-url: https://plot.ly
-BASE_URL: https://plot.ly
-staticurl: https://plot.ly/gh-pages/documentation/static/
+url: /
```
nicolaskruchten: should these go away like in the other files?

_config.yml
```diff
@@ -3,13 +3,6 @@ plugins:
 - jekyll-sitemap
 - algoliasearch-jekyll
 
-# ---
-# Define URLs
-# ---
-url: https://plot.ly
```
nicolaskruchten: so a search for `site.url` or `site.staticurl` or `site.BASE_URL` yields a few results... can we get rid of those please?

scss/main.scss
jdamiba: I have uninlined both `js-splash-highlight.css` and `tomorrow.css` into separate` scss` files.

_config.yml
jdamiba: Yes, I was able to remove all references to these variables.

_posts/pandas/ipython-notebooks/2015-05-25-ipython-notebooks_index.html
```diff
@@ -0,0 +1,376 @@
+---
+title: Pandas for IPython Notebooks | Examples | Plotly
+name: Plotly for IPython Notebooks
+permalink: pandas/ipandas-notebooks/
```
nicolaskruchten: what's this file and permalink? https://plot.ly/python/ipandas-notebooks/ ?

_includes/layouts/seo.html
```diff
@@ -18,8 +18,8 @@
     <meta name="twitter:title" content="{% if page.name %}{{page.name}}{% else %}plotly{% endif %}"/>
     <meta name="twitter:url" content="{% if page.permalink %}https://help.plot.ly{{page.permalink}}{% endif %}"/>
```
nicolaskruchten: Unrelated but I just noticed this problem: can you replace `https://help.plot.ly` here with `https://plot.ly/` (including the trailing slash) please? Right now on URLs like https://plot.ly/python/maps/ it's rendering as `<meta name="twitter:url" content="https://help.plot.lypython/maps/"/>`

nicolaskruchten:
💃

PR #1548, opened 2019-10-10
removing _old scss folder

jdamiba:


nicolaskruchten:
So you're able to regenerate the CSS from the SCSS without errors? Nothing is importing `_old` ?

nicolaskruchten:
💃 (if so)

jdamiba:
@nicolas Yes, I ran `gulp build` to regenerate the scss. That command ran without error and I also have the same Percy diff.

nicolaskruchten:
good to merge

PR #1549, opened 2019-10-11
Add transforms to full reference

### LINKED ISSUE #1354: Full Reference does not contain transform docs
untitled

I saw the really nice transform feature in plotly traces and was trying to do a filter. In the examples - I saw examples for filtering things like `y > 5` etc.
I wanted to do a categorical filter and the data I wanted to filter using was neither x nor y.
So, I thought of going to the Full Reference to understand all the properties it supports.

When I go to Javascript > Full Reference, I could not see any documentation about the transforms except that it is a list of objects in all the trace types:
```
transforms 
Each object has one or more of the keys listed below.
```

chenyejia:
Closes #1354

# Changes
## plotschema
- Add role `polymorphic_array` to `[trace].transforms` to denote that items in array are one of a few types of objects
- Remove `attributes` level from `transforms` and add role `object`
- Add line breaks between operators in `transforms.filter.operations`

## reference-block
- Remove `"transforms"` from list of keys to skip
- Add text describing polymorphic arrays
- Hard-code attribute to render `transforms`
- Use `newline_to_br` filter to process line breaks

# Note
The Algolia reference search should be updated before this is deployed

### DIFF
--- a/_data/plotschema.json
+++ b/_data/plotschema.json
@@ -2233,14 +2233,9 @@
           "description": "Sets the calendar system to use with `y` date data."
         },
         "transforms": {
-          "items": {
-            "transform": {
-              "editType": "calc",
-              "description": "An array of operations that manipulate the trace data, for example filtering or sorting the data arrays.",
-              "role": "object"
-            }
-          },
-          "role": "object"
+          "items": true,
+          "valType": "info_array",
+          "role": "polymorphic_array"
         },
         "uirevision": {
           "valType": "any",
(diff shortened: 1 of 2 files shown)


nicolaskruchten:
Thanks for this PR! Unfortunately, we can't make edits to the plotschema directly, as it's the output of the `getplotschema.py` script, which fetches it from the plotly.js repo.

You're right that the `transforms` attributes are kind of just dangling in place right now, but they're also a semi-deprecated part of our library... I wonder if it wouldn't be better to just remove those keys from the schema reference instead?

Are you making heavy use of `transforms`?

chenyejia:
Ohh shoot, I didn't realize! I'm specifically using groupby a lot, and I wanted to see which filter operations were available.

nicolaskruchten:
Thanks again for this PR. I'm going to close it, as we're shuffling repositories shortly, but when I look at #1354 again more closely I'll likely recycle some of it and credit your work.

Issue #1538, opened 2019-10-02
Standardize <title> tags as a function of name


nicolaskruchten:
Right now our pages' front-matter contains `name` and `title` but I'm 90% sure we can simply generate the `<title>` tag with something like `Plotly | {lang} | {name}` and remove all the `title`s. Let's investigate this.

PR #1562, opened 2019-10-22
Autogenerate Title Tag from Name Front Matter

jdamiba:
closes https://github.com/plotly/documentation/issues/1538

The goal of this PR is to automate the generation of a `title` tag for every post in this repo that isn't just a `redirect` to another post. 

This will allow us to not have to worry about including a `title` tag in the front-matter of posts when writing documentation in the future. 

The format for the `title` tag (which is generated in the `seo.html` `include`) will now be `Post Name | Language | Plotly`.

To enforce this, the script [front-matter-ci.py](https://github.com/plotly/documentation/blob/45739e55c7d5073f41ce4b88a37f905e51af47a4/front-matter-ci.py) has been added and will run in CircleCI and fail  if any post that isn't a `redirect` doesn't have a `name` tag in the front-matter. 

Steps Taken:

- [x] making sure that every post that isn't just a `redirect` has a `name` (see `front-matter-ci.py`)
- [x] editing `seo.html` include to use `name` to generate `title`
- [x] editing `404.html` to have the title fit new format
- [x] removing unneeded `title` front-matter from posts


front-matter-ci.py
```diff
@@ -1,21 +1,35 @@
...
+        except:
+            try:
+                name = post.metadata['jupyter']['plotly']['name']
+            except:
+                try:
+                    if post.metadata['redirect_to']:
+                        continue
+                except:
+                    print(post.metadata['permalink'])
+                    print("CI Check #1 Failed: above post is not a redirect but is missing a name frontmatter")
```
michaelbabyn: Is this still going to make the CI check fail? You raised an Exception before but now this is just a print statement.

It would be nice to continue to collect all the `no_name_paths` here in a list and then raise the exception after the for loop `if len(no_name_paths) > 0`. This way you could also print all those paths at the same time as raising the exception to make debugging easier.

_includes/layouts/seo.html
```diff
@@ -1,6 +1,6 @@
 <!-- SEO Tags - title, meta_description -->
-    <title>{% if page.title != "" or page.title != nil %}{{page.title}}{% elsif page.name != "" or page.name != nil%}{{page.name}}{% else %} Plotly | Online Graphing for Everyone{% endif %}</title>
-    <meta name="description" content="{% if page.description %}{{page.description}}{% else %}Plotly is the easiest way to graph and share your data.{% endif %}">
+<title>{% if page.name %}{{page.name}} | {% if page.language == 'plotly_js' %} JavaScript {% else %}{{page.language | capitalize }}{% endif %} | Plotly {% else %} Plotly Open Source Graphing Libraries{% endif %} </title>
```
michaelbabyn: This looks good

nicolaskruchten:
Ah, I wish you'd held off on merging this until the corresponding PR was done in the plotly.py-docs repo... can you get that PR up ASAP please so we can continue deploying the py docs?

PR #1567, opened 2019-10-23
Add hovertemplate attr to trace

hone1er:
Added a hovertemplate attribute to the first trace so the number shown is equal to the base instead of 0 on hover. 'y' could be used in place of 'base' in the hovertemplate if user wants the number shown to be positive. Seems the user would want either 'y' or 'base' instead of 0 more often than not.


nicolaskruchten:
Thanks!

PR #1568, opened 2019-10-24
Remove has_thumbnail front-matter

jdamiba:
The goal of this PR is to allow documentation writers to not have to worry about including the `has_thumbnail` tag in front-matter in the future. 

Currently, you have to include a `thumbnail` tag and set `has_thumbnail` to `true`. 

Going forward, you will only have to include the hyperlink to the thumbnail, as the layouts simply check for the presence of a `thumbnail` tag.


nicolaskruchten:
Ah, hah, never mind... the `has_thumbnail: true` which I saw removed in that other PR were duplicated :) i.e. python/v3/tree-plots

nicolaskruchten:
For this one:

1. PR to make changes to _layouts and _includes to no longer depend on `has_thumbnail` (and remove `has_thumbnail` from here!)
2. PR in py-docs to remove `has_thumbnail` (which is now safe! because 1 is merged)


No need to have front-matter checks on that.

PR #1572, opened 2019-10-30
using variable for path so front-matter-ci.py can also run in py-docs

jdamiba:
The purpose of this PR is to modify `front-matter-ci.py` so that the path that contains the files to be checked is passed in as a variable rather than hard-coded. 

This is to enable this same script to be run in the `py-docs` repo, reducing duplication.


nicolaskruchten:
💃

PR #1573, opened 2019-10-30
Add title check to front-matter-ci.py

jdamiba:
The purpose of this PR is to ensure that no posts have a `title` front-matter since that is being autogenerated from the `name` front-matter now in the `seo.html` include.


nicolaskruchten:
:dancer: when passes

PR #1584, opened 2019-11-05
Re-order posts in /javascript and /r

jdamiba:
The purpose of this PR is to re-order posts in the `javascript` and `r` docs so that they are sequential integers that start at 1 with no duplicates.  

`check-order.py` and `enforce-order.py` are new utility scripts for keeping the order in a good state.


_posts/plotly_js/fundamentals/responsive-fluid-layout/2015-10-20-responsive-fluid-layout.html
```diff
@@ -1,51 +0,0 @@
----
```
nicolaskruchten: why did this file get deleted?

enforce-order.py
```diff
@@ -0,0 +1,46 @@
...
+try:
+    path = str(sys.argv[1])
+except:
+    raise Exception("You need to specify a path!")
+
+# will contain all posts with display_as: file_settings
+postFamily = []
+  
+#get all posts with frontmatter in md format
+for md_path in Path(path).glob("**/*.html"):
```
nicolaskruchten: can you combine this loop and the next one into one please?

check-order.py
```diff
@@ -0,0 +1,65 @@
...
+
+try:
+    path = str(sys.argv[1])
+except:
+    raise Exception("You need to specify a path!")
+
+# will contain all posts with display_as: file_settings
+postFamily = []
+  
+#get all posts with frontmatter in md format
```
nicolaskruchten: can you combine this loop and the next one into one please?

you can just add all the paths to one combined list and iterate once over the combined list

nicolaskruchten: in fact, the front-matter-ci script should be refactored this way also please

_posts/plotly_js/fundamentals/responsive-fluid-layout/2015-10-20-responsive-fluid-layout.html
jdamiba: This file was deleted because it is not being served in production at https://plot.ly/javascript/responsive-fluid-layout/

jdamiba:
@nicolaskruchten This PR has been updated to combine the loops to collect the `html` and `md` format posts into one loop.

check-order.py
```diff
@@ -0,0 +1,50 @@
...
+for path in paths:
+    post = frontmatter.load(str(path))
+    if len(post.metadata.keys()) > 0:
+        if "display_as" in post.metadata:
+            if post.metadata['display_as'] == category:
+                postFamily.append({'path':str(path), 'order' : post.metadata['order']})
+        
+sortedPostFamily = sorted(postFamily, key = lambda i: i['order'])
+
+order = []
```
nicolaskruchten: this little block could be a comprehension: `order = [ p['order'] for p in sortedPostFamily ]`

enforce-order.py
```diff
@@ -0,0 +1,42 @@
...
+for path in paths:
+    post = frontmatter.load(str(path))
+    if len(post.metadata.keys()) > 0:
+        if "display_as" in post.metadata:
+            if post.metadata['display_as'] == category:
+                postFamily.append({'path':str(path), 'order' : post.metadata['order']})
+        
+sortedPostFamily = sorted(postFamily, key = lambda i: i['order'])
+
+def enforceOrder(listToBeOrdered):
```
nicolaskruchten: this script shares so much with the one above that they should be combined, with an extra command-line argument for "enforce"

check-order.py
```diff
@@ -0,0 +1,50 @@
...
+for post in sortedPostFamily:
+   order.append(post['order'])
+
+print(order)
+
+if order[0] != 1:
+    raise Exception("Order Check Failed! First post does not have order 1!")
+
+def checkConsecutive(l): 
+    return sorted(l) == list(range(min(l), max(l)+1))
```
nicolaskruchten: this could just be `sorted(l) == list(range(1, len(l)+1)` instead, right? we want things to start at 1

nicolaskruchten: that would let you remove the `if` above

PR #1593, opened 2019-11-13
re-order posts in /python-v3, /r, and /javascript docs

jdamiba:
The purpose of this PR is to re-organize the `order` front-matter in `python-v3`, `plotly_js`, and `r` documentation posts. From now on, the order within a top 8 `display_as` category will start at 1 and increase sequentially using integers.


check-or-enforce-order.py
```diff
@@ -0,0 +1,74 @@
+import frontmatter
+from pathlib import Path, PosixPath
+import sys
+
+try:
```
nicolaskruchten: let's make it clear that the path here is intended to include only posts from a single language

check-or-enforce-order.py
```diff
@@ -0,0 +1,74 @@
...
+            postToBeAltered = frontmatter.load(post['path'])
+            postToBeAltered.metadata['order'] = index+1
+            frontmatter.dump(postToBeAltered, post['path'])
+
+def checkConsecutive(l): 
+        return sorted(l) == list(range(1, len(l)+1)) 
+
+categories = ["file_settings", "basic", "statistical", "scientific", "maps", "3d_charts", "multiple_axes"]
+
+try:
```
nicolaskruchten: this second check can go

check-or-enforce-order.py
```diff
@@ -0,0 +1,74 @@
...
+                if post.metadata['display_as'] == category:
+                    postFamily.append({'path':str(md_path), 'order' : post.metadata['order']})
+                
+    sortedPostFamily = sorted(postFamily, key = lambda i: i['order'])
+
+    order = [ p['order'] for p in sortedPostFamily ]
+
+    print(order)
+
+    if order[0] != 1:
```
nicolaskruchten: i think this block is unnecessary

check-or-enforce-order.py
```diff
@@ -0,0 +1,74 @@
...
+
+    if not checkConsecutive(order):
+        if len(sys.argv) == 3:
+            if sys.argv[2] == 'enforce':
+                print('Order Check Did Not Pass! ENFORCING CORRECT ORDER for {}'.format(category))
+                enforceOrder(sortedPostFamily)
+        else: 
+            raise Exception("Order Check Failed! Order front-matter in '{}' display_as are not consecutive integers!!".format(category))
+
+    if len(order) != len(set(order)):
```
nicolaskruchten: let's try to roll this into `checkConsecutive` somehow and only execute that try-catch once

check-or-enforce-order.py
```diff
@@ -0,0 +1,74 @@
...
+    for md_path in paths:
+        post = frontmatter.load(str(md_path))
+        if len(post.metadata.keys()) > 0:
+            if "display_as" in post.metadata:
+                if post.metadata['display_as'] == category:
+                    postFamily.append({'path':str(md_path), 'order' : post.metadata['order']})
+                
+    sortedPostFamily = sorted(postFamily, key = lambda i: i['order'])
+
+    order = [ p['order'] for p in sortedPostFamily ]
```
nicolaskruchten: we should be able to use one script by conditionally adding a 5 for when the path is python

check-or-enforce-order.py
```diff
@@ -0,0 +1,74 @@
...
+    file_path = str(sys.argv[1])
+except:
+    raise Exception("You need to specify a path!")
+
+def enforceOrder(listToBeOrdered):
+    for index, post in enumerate(listToBeOrdered):
+        if index+1 != post['order']:      
+            postToBeAltered = frontmatter.load(post['path'])
+            postToBeAltered.metadata['order'] = index+1
+            frontmatter.dump(postToBeAltered, post['path'])
```
nicolaskruchten: here in python you'll have to skip 5

check-or-enforce-order.py
```diff
@@ -0,0 +1,65 @@
...
+# path here is intended to include only posts from a single language
+try:
+    file_path = str(sys.argv[1])
+except:
+    raise Exception("You need to specify a path!")
+
+def enforceOrder(listToBeOrdered):
+    for index, post in enumerate(listToBeOrdered):
+        if file_path == 'build/html':
+            if index+1 == 5:
```
nicolaskruchten: so I think this will not work, you end up skipping one post...

check-or-enforce-order.py
```diff
@@ -0,0 +1,57 @@
...
+# path here is intended to include only posts from a single language
+try:
+    file_path = str(sys.argv[1])
+except:
+    raise Exception("You need to specify a path!")
+
+def enforceOrder(listToBeOrdered):
+    for index, post in enumerate(listToBeOrdered):    
+        postToBeAltered = frontmatter.load(post['path'])
+        postToBeAltered.metadata['order'] = (index+1 if file_path != "build/html" and index<5 else index+2)
```
nicolaskruchten: I think this condition I banged out in Slack isn't quite right actually... it'll add 2 for indices greater than 5 even outside of build/html. sorry :)

nicolaskruchten: it should be `(index+2 if file_path == "build/html" and index>=5 else index+1)` I think

check-or-enforce-order.py
```diff
@@ -0,0 +1,57 @@
...
+        frontmatter.dump(postToBeAltered, post['path'])
+
+def checkConsecutive(l): 
+    if len(order) != len(set(order)):
```
nicolaskruchten: I'm pretty sure this check is redundant, i.e. that anything that fails this will fail the next one.

This function could in principle boil down to simply `return sorted(l) == list(range(1, len(l)+1))` except that this would fail in the `py-docs` repo because of the missing 5, so  it probably needs to be something like:

```python
if file_path == "build/html":
    l = l + [5]
return sorted(l) == list(range(1, len(l)+1))
```

PS: having a single-character variable `l` makes it hard to read expressions that add `1` in many fonts ;)

check-or-enforce-order.py
jdamiba: Updated the function with the suggested refactor.

check-or-enforce-order.py
jdamiba: Updated the method with the suggested refactor and made the variable name more explicit to reduce confusion.

check-or-enforce-order.py
```diff
@@ -0,0 +1,95 @@
...
+                frontmatter.dump(postToBeAltered, post['path'])
+                rPath = post['path'][:-3] + '.Rmd'
+                try: 
+                    rPostToBeAltered = frontmatter.load(rPath)
+                    rPostToBeAltered.metadata['order'] = index+1
+                    frontmatter.dump(rPostToBeAltered, rPath)
+                except:
+                    continue
+        elif file_path == "python": # accounts for the fact that this is also run in the plotly.py-docs repo
+            if post['path'][0] == ".":
```
nicolaskruchten: you can skip this check (see comment below)

check-or-enforce-order.py
```diff
@@ -0,0 +1,95 @@
...
+    return sorted(listToBeChecked) == list(range(1, len(listToBeChecked)+1))
+
+def main():
+    # 1. collect the current order of posts 
+    # 2. sort and check if sorted order is sequential
+    for category in categories:
+        postFamily = []
+        #get all posts with frontmatter in md format
+        for md_path in paths:
+            post = frontmatter.load(str(md_path))
```
nicolaskruchten: here you should add:
 
if ".ipynb_checkpoints" in str(md_path):
    continue
```

File: check-or-enforce-order.py [hunk 12]
```diff
@@ -0,0 +1,95 @@
...
+        return post["jupyter"]["plotly"]
+    else:
+        return post.metadata
+
+# this function will mutate the front-matter to enforce a sequential order
+def enforceOrder(listToBeOrdered):
+    for index, post in enumerate(listToBeOrdered):
+        postToBeAltered = frontmatter.load(post['path'])
+        if file_path == "_posts/r": # accounts for the fact that sometimes there are both .md and .Rmd files
+            if post['path'][-3:] == ".md":
```
nicolaskruchten: what if the file is an HTML file? do we just not touch the order there or?

check-or-enforce-order.py
jdamiba: Skipped.

check-or-enforce-order.py
jdamiba: Added.

check-or-enforce-order.py
jdamiba: Currently, there aren't any HTML files with `order` front-matter in the R documentation.

nicolaskruchten:
💃

PR #1598, opened 2019-11-15
refactor ci scripts

jdamiba:
The purpose of this PR is to refactor the scripts that run in CircleCI for performance and to add additional checks. 

- [x] fix typo in `front-matter-ci.py` => `hmtl` to `html`
- [x] add check to make sure that no post with `order` greater than 5 also has `page_type: example_index`
- [x] add check to make sure that all posts in Python, R, and JavaScript in the top 8 categories have a `thumbnail` front-matter.
- [x] refactor `check-or-enforce-order.py` to reduce time and space complexity of algorithm
- [x] add check to make sure permalinks have a trailing slash + fixup to posts missing trailing slash
- [x] fixup to posts with extraneous redirect_froms
- [x] only raise `Exception` after printing results of all tests in `front-matter-ci.py` and `check-or-enforce-order.py`
- [x] under_score variable names rather than camelCase

### DIFF
--- a/check-or-enforce-order.py
+++ b/check-or-enforce-order.py
@@ -21,15 +21,6 @@ categories = ["file_settings", "basic", "financial", "statistical", "scientific"
 def get_post(path):
     return fm.load(str(path))
 
-def get_all_posts(paths):
-    posts = []
-    for path in paths:
-        post = get_post(path)
-        front_matter = get_front_matter(post)
-        if validate_front_matter(front_matter):
-            posts.append({'path':path, 'order' : front_matter['order'], 'display_as' : front_matter['display_as']})
-    return posts
-
 def get_front_matter(post):
     if "jupyter" in post.metadata:
         return post["jupyter"]["plotly"]
--- a/check-or-enforce-order.py
+++ b/check-or-enforce-order.py
@@ -1,95 +1,120 @@
-import frontmatter
+import frontmatter as fm
 from pathlib import Path, PosixPath
 import sys
 
 # path here is intended to include only posts from a single language
+# _posts/r, _posts/plotly_js, _posts/python-v3, _posts/python in 'documentation'
+# build/html in 'plotly.py-docs'
 try:
     file_path = str(sys.argv[1])
 except:
     raise Exception("You need to specify a path!")
 
-# check to see if enforce flag was given
+# check to see if enforce flag was given at command line
 enforce = False
 if len(sys.argv) == 3:
     if sys.argv[2] == 'enforce':
         enforce = True
 
-# post families with these strings as "display_as" front-matter will be checked
 categories = ["file_settings", "basic", "financial", "statistical", "scientific", "maps", "3d_charts", "multiple_axes"]
 
-paths = []
-for suffix in ["md", "html"]:
-    paths += [x for x in Path(file_path).glob("**/*."+suffix)]
+def get_post(path):
+    return fm.load(str(path))
 
-def get_meta(post):
+def get_all_posts(paths):
+    posts = []
+    for path in paths:
+        post = get_post(path)
+        front_matter = get_front_matter(post)
+        if validate_front_matter(front_matter):
+            posts.append({'path':path, 'order' : front_matter['order'], 'display_as' : front_matter['display_as']})
+    return posts
+
+def get_front_matter(post):
     if "jupyter" in post.metadata:
         return post["jupyter"]["plotly"]
     else:
         return post.metadata
 
 # this function will mutate the front-matter to enforce a sequential order
-def enforceOrder(listToBeOrdered):
-    for index, post in enumerate(listToBeOrdered):
-        postToBeAltered = frontmatter.load(post['path'])
+def enforceOrder(list_to_be_ordered):
+    print(list_to_be_ordered)
+    for index, post in enumerate(list_to_be_ordered):
+        post_to_be_altered = fm.load(post)
         if file_path == "_posts/r": # accounts for the fact that sometimes there are both .md and .Rmd files
-            if post['path'][-3:] == ".md":
-                postToBeAltered.metadata['order'] = index+1
-                frontmatter.dump(postToBeAltered, post['path'])
-                rPath = post['path'][:-3] + '.Rmd'
+            if post[-3:] == ".md":
+                post_to_be_altered.metadata['order'] = index+1
+                fm.dump(post_to_be_altered, post)
+                rPath = post[:-3] + '.Rmd'
                 try: 
-                    rPostToBeAltered = frontmatter.load(rPath)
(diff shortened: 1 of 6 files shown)


check-or-enforce-order.py
```diff
@@ -1,27 +1,34 @@
...
-    paths += [x for x in Path(file_path).glob("**/*."+suffix)]
+def get_paths():
+    paths = []
+    for suffix in ["md", "html"]:
+        paths += [x for x in Path(file_path).glob("**/*."+suffix)]
+    return paths
 
-def get_meta(post):
+def get_post(path):
+    if ".ipynb_checkpoints" not in str(path):
```
nicolaskruchten: It's an anti-pattern to conditionally return something from a function... what happens "else" here? Nothing is returned? won't that cause an exception downstream? you should probably raise an exception here and wrap your call below in a try/except that does `continue` in the `except` or something

check-or-enforce-order.py
```diff
@@ -1,27 +1,34 @@
...
+        return fm.load(str(path))
+
+def get_front_matter(post):
```
nicolaskruchten: this is an example of a "good" pattern where the return type is always the same (compared to above)

check-or-enforce-order.py
```diff
@@ -30,62 +37,79 @@ def get_meta(post):
...
-                if "display_as" in metadata:
-                    if metadata['display_as'] == category:
-                        postFamily.append({'path':str(md_path), 'order' : metadata['order']})
-        
-        sortedPostFamily = sorted(postFamily, key = lambda i: i['order'])
+        order = []
+        sortedCategory = []
 
-        order = [ p['order'] for p in sortedPostFamily ]
+        for post in sortedPosts:
```
nicolaskruchten: this is still a nested loop, right? so there's no time-complexity benefit to this refactoring yet...?

front-matter-ci.py
```diff
@@ -2,73 +2,96 @@
...
+        duplicatePermalinks.append(post)
         continue
     else:
         temp.append(post)
 
-if len(dupeLinks) > 0:
-    raise Exception("CI Check #3 Not Passed: Following permalinks:\n{} are duplicated!\n".format('\n'.join([str(item) for item in dupeLinks])))
-print("CI Check #3 Passed: No duplicate permalinks!")
\ No newline at end of file
+if len(duplicatePermalinks) > 0:
```
nicolaskruchten: so this set of blocks at the bottom could easily be DRYed up... Basically above you're collecting a list of "violations" and the checks pass if the lists are empty. So you should be able to write a little for-loop that iterates over your violation lists and emits these error messages.

Also, right now if multiple checks don't pass, you only find out about the first one because it raises and errors out. It's always really nice for CI to tell you all the things that are wrong all at once so you can fix them in one shot rather than have to iterate over and over :)

jdamiba:
@nicolaskruchten I have updated this PR in response to your comments, specifically regarding the time complexity of `check-or-enforce-order.py` and DRYing up the checks in `front-matter-ci.py`. 

For the former, you were right that I hadn't addressed the time-complexity issue. I think I've figured out how to use a list comprehension in order to only have to loop through the list of paths once during the program's execution.  

For the latter, rather than using a loop, I defined a new function `ci_check` which accepts as its parameters a list and an error message. I believe this achieves the goals of DRYing up that part of the program and allowing the program to display multiple error messages.

check-or-enforce-order.py
```diff
@@ -30,66 +39,85 @@ def get_meta(post):
...
+            return False
+    else:
+        return False
 
-        order = [ p['order'] for p in sortedPostFamily ]
+def get_all_posts(paths):
+    posts = []
+    for path in paths:
+        try:
+            post = get_post(path)
```
nicolaskruchten: this will never raise an exception, as it just returns `False` above

jdamiba: I removed the anti-pattern by adding the `if ".ipynb_checkpoints" not in str(file_path)` conditional to the list comprehension in `get_paths()`. That way the list of `paths` never contains those files to begin with and I can avoid the extra method call.

check-or-enforce-order.py
```diff
@@ -30,66 +36,81 @@ def get_meta(post):
...
 
-        print(order)
+def get_all_posts(paths):
+    posts = []
+    for path in paths:
+        post = get_post(path)
+        front_matter = get_front_matter(post)
+        if validate_front_matter(front_matter):
+            posts.append({'path':path, 'order' : front_matter['order'], 'display_as' : front_matter['display_as']})
+        else:
```
nicolaskruchten: `else: continue` is not needed here: you're already at the end of the loop :)

check-or-enforce-order.py
```diff
@@ -30,66 +36,81 @@ def get_meta(post):
...
+def main():
+    # 1. collect the current order of posts and check if sorted order is sequential
+    allPaths = get_paths()
+    allPosts = get_all_posts(allPaths)
+    sortedPosts = sorted(allPosts, key = lambda i: (i['display_as'], i['order']))
+    order = []
+    postsToBeOrdered = []
+    
+    for category in categories:
+        order = [post['order'] for post in sortedPosts if post['display_as'] == category]
```
nicolaskruchten: you're still iterating over all posts within a for-each-category loop here :)

check-or-enforce-order.py
```diff
@@ -1,95 +1,120 @@
...
     if sys.argv[2] == 'enforce':
         enforce = True
 
-# post families with these strings as "display_as" front-matter will be checked
 categories = ["file_settings", "basic", "financial", "statistical", "scientific", "maps", "3d_charts", "multiple_axes"]
 
-paths = []
-for suffix in ["md", "html"]:
-    paths += [x for x in Path(file_path).glob("**/*."+suffix)]
+def get_post(path):
```
nicolaskruchten: are these utility functions still relevant? can you do a dead-code-elimination pass plz?

nicolaskruchten:
Can you please open a PR in `py-docs` that runs this branch's version of the scripts, just so we can confirm it's all green?

check-or-enforce-order.py
```diff
@@ -1,95 +1,111 @@
...
+            return False
+    else:
+        return False
 
-        print(order)
+def get_paths_and_orders_by_category():
+    posts_by_category = {category: dict(orders=[], paths=[]) for category in categories}
+    for suffix in ["md", "html"]:
+        for path in Path(file_path).glob("**/*."+suffix): 
+            if ".ipynb_checkpoints" not in str(file_path):
```
nicolaskruchten: this should be `not in str(path)` otherwise enforcement fails in Python

nicolaskruchten:
💃 once `file_path` is changed to `path` on that one line.

Nice job :)

PR #1606, opened 2019-11-21
Fix Escape Errors in Julia Docs

jdamiba:
closes https://github.com/plotly/documentation/issues/1382

The purpose of this PR is to replace `=&gt;` and `=&lt;` with `>` and `<` in order to make the code examples work. 

Before:
Screen Shot 2019-11-22 at 3 38 41 PM

After:
Screen Shot 2019-11-22 at 3 38 59 PM

Embedded charts still work.


nicolaskruchten:
You've spot-checked to make sure these all render correctly etc locally?

_posts/plotly_js/2015-07-21-plotlyjs-getting-started.html
```diff
@@ -14,12 +14,12 @@ <h2>Download</h2>
             </p>
 
             <p>
-                Include the downloaded scripts before the end of the &lt;/head&gt; tag in your HTML document:
+                Include the downloaded scripts before the end of the </head> tag in your HTML document:
```
nicolaskruchten: ah so we need to be pretty careful here... this one is actual HTML so the escaping is desired.

nicolaskruchten: The others seem legit but we'll want to check them I think. Also, some of the auto-example ones below... do they actually run with your new inline system? seems like they wouldn't, as they're not valid JS...

Issue #1587, opened 2019-11-06
inline Javascript examples (no more embedded CodePens)


nicolaskruchten:
JS examples should be rendering plots "in line" in the page so that they always match the displayed code, rather than embedding an external CodePen which easily gets out of date.

jdamiba:
This will require refactoring lines 24-35 of `auto_examples.html` to inline a `script` tag rather than an `iframe` with codepen embed. 

This shouldn't be an issue as the `plotly.js` library is present in the browser environment.

jdamiba:
This issue has been addressed.

PR #1605, opened 2019-11-20
Inline js codepens

jdamiba:
The purpose of this PR is to inline `plotly.js` example charts into this repo rather than relying on `codepen` embeds. 

Most charts were inlined with no issue, however a couple of posts are outliers. 

Charts to remain as embedded codepens:

- [x] /javascript/react - needs React to be loaded clientside
- [x] /javascript/pointcloud - needs canvas/2d-context in the DOM
- [x] /javascript/heatmap-webgl - needs canvas/2d-context in the DOM
- [x] /javascript/2d-density-plots - not codepen, jsfiddle with a chart studio chart
- [x] javascript/disable-zoom/#disabling-zoom-events-for-x-axis - not sure why this chart does't render
- [x] javascript/hover-events/#combined-click-and-hover-events - requires buttons in DOM Structure
- [x] javascript/parallel-categories-diagram/ - expects radio buttons in DOM
- [x] javascript/animations/ uses a lot of buttons.
- [x] javascript/hover-events/ uses a `<div>` to show hover info
- [x] javascript/dropdowns/#bind-dropdown-events-to-plotlyjs-charts uses an HTML dropdown menu
- [x] javascript/uirevision/

PR #1616, opened 2019-12-09
Inline JS example charts and add edit on codepen button

jdamiba:
The purpose of this PR is to inline the example charts in the Plotly.js documentation while preserving the `edit on codepen` functionality. This will allow contributors to write docs without having to worry about maintaining a separate codepen, while users still get the benefit of being able to use/edit live code snippets. 

In order to accomplish this, I

- [x] made sure to import the plotly.js library into the DOM on pages with plotly.js charts
- [x] created an `embed:true` front-matter property so that charts that require special DOM elements can remain as codepen embeds
- [x] edit example code to standardize use of `myDiv` as container div for chart
- [x] updated the Jekyll layout`auto_examples.html` to inline charts
- [x] add an `edit on codepen` button that submits a POST request to dynamically generate the codepen

Before:

Screen Shot 2019-12-09 at 3 35 21 PM

After:
Screen Shot 2019-12-09 at 3 36 30 PM

 closes https://github.com/plotly/documentation/issues/1587


jdamiba:
FYI @nicolas this PR incorporates all the QA from and supersedes https://github.com/plotly/documentation/pull/1605

jdamiba:
@nicolas I have updated the styling of inline charts (removed border) and moved the `try it on codepen` button into the code area. 

Screen Shot 2019-12-12 at 4 27 56 PM

nicolaskruchten:
Could we trim the whitespace above the button a bit and change the font on it to be open sans plz?

_includes/posts/auto_examples.html
```diff
@@ -12,7 +14,14 @@ <h4 id="{{ example.name | replace:' ','-' | remove:',' | downcase | strip_html }
           {% endif %}
           {% if example.content != "" %}
             <div class="z-depth-1">
-              <pre><code class="{{page.language}}">{{ example.content | xml_escape }}</code></pre>
+              <pre>
```
nicolaskruchten: so the `<pre>` tag here is very sensitive to whitespace, which is why the `<code>` tag was stuck next to it before, and which is why there is so much extra whitespace now. Let's bring the old line back and move the `<form>` outside the `<pre>` so the button is immediately under the grey box, in serif font, and we trim the whitespace before and after the code back to what it used to be.

nicolaskruchten:
The React page is a bit wonky, because there is the "try it on codepen" link on all the examples, even the ones that don't have an embed and don't run, like the `npm install` block, and the ones that do. We should not show the "try it" button in these cases

nicolaskruchten:
Some of these examples don't work because they try to write to an non-`myDiv` div: http://localhost:4000/javascript/hover-events/

nicolaskruchten:
The top one here needs the `myDiv` conversion: http://localhost:4000/javascript/disable-zoom/

nicolaskruchten:
Actually there are quite a few places where we need to suppress the "try it" button, e.g. at the top of http://localhost:4000/javascript/tick-formatting/

Maybe we can just show it if there is `myDiv` in the string? that's probably a decent heuristic

nicolaskruchten:
This one doesn't work because it uses `require` http://localhost:4000/javascript/colorway/ ... we should ensure that anything that uses `require` either gets a workaround or goes back to being embedded in a pen.

nicolaskruchten:
This is getting super, super close! Let's clear up the last few comments I made here and we can merge this this week :)

nicolaskruchten:
Even if we rely on a bunch of CodePens for now, the mere possibility of creating new docs without them is great, and the fact that most of our examples are not using CodePen directly any more is a huge win.

_includes/posts/auto_examples.html
```diff
@@ -23,18 +32,33 @@ <h4 id="{{ example.name | replace:' ','-' | remove:',' | downcase | strip_html }
...
         {% if example.plot_url %}
           <div class="{% if example.arrangement contains 'horizontal' %}twelve{% else %}six{% endif %} columns">
-            <div class="border" style="text-align:center;">
               {% if example.plot_url contains '.gif' or example.plot_url contains '.png' %}
               <img src="{{ example.plot_url }}" />
               {% else %}
-              <iframe id="auto-examples" src="{{ example.plot_url }}{% if example.plot_url contains 'plot.ly' %}.embed{% endif %}"
-                style="width: {% if example.width %}{{example.width}}px;{% else %}100%;{% endif %} height: {% if example.height %}{{example.height}}px;{% else %}550px;{% endif %} border: none;"></iframe>
+                {% if page.language == "plotly_js"%}
+                  {% if example.embed %}
```
nicolaskruchten: As a pattern, could we instead not use the `embed` flag and say "if there is a `plot_url` embed it, otherwise try to render inline"?

PR #1618, opened 2019-12-16
basic examples without codepens

jdamiba:


_includes/posts/auto_examples.html
```diff
@@ -23,18 +26,29 @@ <h4 id="{{ example.name | replace:' ','-' | remove:',' | downcase | strip_html }
         </div>
         {% if example.plot_url %}
           <div class="{% if example.arrangement contains 'horizontal' %}twelve{% else %}six{% endif %} columns">
-            <div class="border" style="text-align:center;">
```
nicolaskruchten: is there a matching `</div>` that needs to be removed here?

jdamiba: Yes, removed.

nicolaskruchten:
Epic! 💃

PR #1623, opened 2019-12-16
rest of examples removing plot-url

jdamiba:


nicolaskruchten:
bottom example of the streamtubes page is not looking right

nicolaskruchten:
OK let's log an issue for the bottom streamtube example and merge.

💃

PR #1635, opened 2019-12-17
use r-docs-branch to pull many R examples

jdamiba:


.circleci/config.yml
```diff
@@ -54,7 +54,9 @@ jobs:
             git config --global user.email "accounts@plot.ly"
             git config --global user.name "Deployer"
             rm -rf _posts/python/html
+            rm -rf _posts/r/basic
```
nicolaskruchten: this should be `r/md` as below :)

nicolaskruchten:
OK, let's do this. 💃

PR #1649, opened 2020-01-07
fix legend-inside by specifying xanchor to right

antoinerg:
Right now the "legend inside" example is broken:
newplot (38)

This PR fixes it:
newplot (39)


jdamiba:
Percy is just complaining about the usual set of false positives. 

LGTM 💃

PR #1655, opened 2020-01-13
replace logo in dash ad

jdamiba:
closes https://github.com/plotly/marketing-team/issues/193

### DIFF
--- a/_includes/layouts/helpbox.html
+++ b/_includes/layouts/helpbox.html
@@ -7,7 +7,7 @@
         <img
           height="50%"
           width="50%"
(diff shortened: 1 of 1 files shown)


jdamiba:
Please note that if you are hovering over a Percy diff it adds a white border around the image.

nicolaskruchten:
Looks good but the new image is enormous (8000 pixels wide!) and so will slow down the page a lot. Can you resize it to be max 1000px wide or something please?

jdamiba:
I have resized the `dash-ad-2` image and re-uploaded it to S3. I also reran Percy so the current diffs are with the newly resized image.

PR #1662, opened 2020-01-16
Delete r/chart-studio files that are moving to r-docs

jdamiba:
The purpose of this PR is to delete files in the `r/chart-studio` folder which are being moved over to `plotly.r-docs`.


nicolaskruchten:
Ah, I don't think these files should move to the r-docs repo... we don't want to run them on every commit, as they communicate with Chart Studio.

jdamiba:
Can you elaborate on why it would not be a good idea to run these files on every commit? Is it because charts would accumulate in the Chart Studio account that is linked to this repo? FYI, these files pass CI checks on the feature branch in the other repo.

nicolaskruchten:
I just don't want to have these doc builds dependent on external services, and we've seen issues with updating/overwriting plots simultaneously on multiple concurrent CI runs in Python-land.

nicolaskruchten:
This can be closed, right?

jdamiba:
Superseded by https://github.com/plotly/documentation/pull/1663

PR #1666, opened 2020-01-22
update ci script

jdamiba:
The purpose of this PR is to update the CI script so that it will highlight posts with `order` less than five but also have `page_type: u-guide`.  This combination will not cause the build to fail but will prevent the post from being on the language index. 

This PR will fail CI until https://github.com/plotly/plotly.py/pull/2115 is merged, as those posts currently fail this test.


nicolaskruchten:
💃 once all is green again

PR #1668, opened 2020-01-24
update modebar docs

jdamiba:
The purpose of this PR is twofold:

First, to update some of the language around modebar configuration to make more explicit what options users have to add or remove icons. 
- [x] delete old examples
- [x] write new examples

Second, to bring this doc page in line with other pages in terms of having example figures generated inline rather than using embedded codepens. 
- [x] remove `plot_url` front-matter to inline chart
- [x] make sure all figures render appropriately

Thanks @Mahdis-z for pointing this out!


_posts/plotly_js/fundamentals/config-options/2020-01-24-display-modebar.html
```diff
@@ -0,0 +1,21 @@
+---
+name: Display Modebar
```
nicolaskruchten: instead of "hide" and "display" the modebar can we find some way of saying "Disable modebar" and "Force modebar to be always visible" with an explanation of the default behaviour and what this override does?

I say this because the modebar already does "hide" and "display" by default, so just saying "hide" and "display" isn't quite clear enough.

nicolaskruchten: We could also just restore the old title with the "always" for now and fix it later :)

nicolaskruchten:
:dancer: once we address the "always" comment

